PR #1, opened 2017-02-27
Breaking: rename CollapseElement to CollapseTable

niedzielski:
Try to limit scope by renaming CollapseElement to CollapseTable


niedzielski:
@montehurd, cool, I've pushed v1.0.0!

PR #2, opened 2017-03-16
Remove "build" folder and contents from git tracking

montehurd:
I don't think we want to track build folder contents. 

When "npm install" fetches this package it implicitly builds: https://docs.npmjs.com/cli/build 

This will make it easier to open PR's :)


niedzielski:
@montehurd, thanks for schooling me! This looks like a great improvement! [This lifecycles page](https://docs.npmjs.com/misc/scripts) feels like it should mention both build and prebuild scripts.

PR #3, opened 2017-03-17
New: CollapseTable

niedzielski:


test/CollapseTable.js
```diff
@@ -48,4 +48,104 @@ describe('CollapseTable', () => {
...
+                            style='visibility: hidden'
+                          </span>
+                        </div>
+                        <table style='display: block'></table>
+                        <div id=footer style='display: block'></div>
+                      </div>`
+        this.doc = domino.createDocument(html)
+      })
+
+      it('where the footer is observed, the callback is invoked when the footer is clicked',
```
niedzielski: I tried a more descriptive style for each test case. Let me know your thoughts on this and I'll make the tests consistent one way or another.

montehurd: I like the descriptive style! 👍

niedzielski:
@montehurd, ready for review with:

https://gerrit.wikimedia.org/r/#/c/349371/
https://gerrit.wikimedia.org/r/#/c/349373/

src/CollapseTable.css
```diff
@@ -0,0 +1,140 @@
...
+  line-height: 120%;
+  background-color: #f8f8f8;
+  background-repeat: no-repeat;
+  background-position: 95% 50%;
+  background-size: 16px 16px;
+}
+
+/* The outermost table container header's subcaption text. todo: rename. */
+.app_span_collapse_text {
+  color: #808080;
```
niedzielski: These colors differ from iOS. I'm not sure any of these styles should be here in the long term (maybe just empty classes or classes with visibility changes).

montehurd:
Made a quick pass - will poke around more tomorrow! Exciting stuff!!!

src/transform/CollapseTable.css
```diff
@@ -0,0 +1,140 @@
...
+.content table.infobox {
+  font-size: 100%;
+  border-style: none !important;
+}
+
+table.infobox caption {
+  text-align: center;
+  font-weight: bold;
+}
+
```
montehurd: Do we need the `figcaption` class I see [here](https://gerrit.wikimedia.org/r/#/c/349373/1/app/src/main/assets/applib.css)?

niedzielski: I think this is now here: https://github.com/wikimedia/wikimedia-page-library/blob/master/src/transform/WidenImage.css#L41

montehurd: Ah that's right!

src/transform/CollapseTable.js
```diff
@@ -48,6 +51,182 @@ const getTableHeader = (element, pageTitle) => {
...
+    table.style.display = 'none'
+    header.classList.remove('app_table_collapse_close') // todo: use app_table_collapsed_collapsed
+    header.classList.remove('app_table_collapse_icon') // todo: use app_table_collapsed_icon
+    header.classList.add('app_table_collapsed_open') // todo: use app_table_collapsed_expanded
+    if (caption) {
+      caption.style.visibility = 'visible'
+    }
+    footer.style.display = 'none'
+        // if they clicked the bottom div, then scroll back up to the top of the table.
+    if (this === footer && toggleCollapseClickCallback) {
```
montehurd: I think `toggleCollapseClickCallback` should be `footerDivClickCallback` here?

src/transform/CollapseTable.js
```diff
@@ -48,6 +51,182 @@ const getTableHeader = (element, pageTitle) => {
...
+  for (let i = 0; i < tables.length; ++i) {
+    const table = tables[i]
+
+    if (elementUtilities.findClosestAncestor(table, '.app_table_container')
+      || !shouldTableBeCollapsed(table)) {
+      continue
+    }
+
+    // todo: this is actually an array
+    const headerText = getTableHeader(table, pageTitle)
```
montehurd: Do you have an example of a page where the transform ends up using `pageTitle`?

montehurd: I don't see that parameter in the iOS transform so just curious.

montehurd:
We seem to be missing `[openCollapsedTableIfItContainsElement](https://github.com/wikimedia/wikipedia-ios/blob/develop/www/js/transforms/collapseTables.js#L159)`.

We use it on iOS if a [link target is inside a closed table](https://github.com/wikimedia/wikipedia-ios/blob/develop/www/js/listeners.js#L54).

I'll look for an example...

src/transform/CollapseTable.js
niedzielski: Yes! Fixed and added a test.

src/transform/CollapseTable.js
niedzielski: iOS uses window.pageTitle directly. This should now be passed as a parameter. I believe a collapsed references section (third example in CollapseTable.css) would be an example use.

niedzielski:
@montehurd all comments addressed.

> openCollapsedTableIfItContainsElement

Thank you! I'm not sure how I missed this. I've added the method with some tests, however...
- Name changes to **expand**CollapsedTableIfItContainsElement() for consistency (I don't care if we use open and close or expand and collapse but I do care about consistency!)
- The implementation differs from the original in that ParentNode.firstElementChild is used rather than Node.firstChild since classList is only available on Elements.

src/transform/CollapseTable.js
```diff
@@ -218,9 +218,27 @@ const collapseTables = (document, content, pageTitle, isMainPage, infoboxTitle,
   }
 }
 
+/**
+ * @param  {?Element} element
+ * @return {void}
```
montehurd: Wondering how we want to deal with things like a comment reminding us where issues can be seen... i.e. for this one:
> See first reference link - the `[A]` in `enwiki > Airplane` article from ~June 2016.

src/transform/CollapseTable.js
montehurd: Hmm I think I'm missing something... if I override this parameter with `ASDFASFDASDFA` I don't seem to be able to find any collapsed or expanded table which shows the `ASDFASFDASDFA` string... I've tried quite a few articles...

montehurd:
`expandCollapsedTableIfItContainsElement` works great!

montehurd:
Oops sorry hit the wrong button!

montehurd:
Good news! The iOS integration looks pretty good: https://github.com/wikimedia/wikipedia-ios/pull/1403

src/transform/CollapseTable.js
niedzielski: Resolved offline + adding some comments.

niedzielski:
:+1: Examples, todos, and comments added. All comments addressed.

montehurd:
THIS IS GREAT!

PR #6, opened 2017-03-22
Add "rollup-plugin-css-porter" so applib can deliver css too.

montehurd:
For instance, the "WidenImage.js" transform can then also have a "WidenImage.css" file and NPM will not only build "applib.js" but also "applib.css".

Respective iOS and Android "Gruntfile.js" will need be modified once to digest "node_modules/applib/build/applib.css" in platform appropriate way (to get "applib.css" into respective "assets" directories).

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,10 +1,11 @@
 import babel from 'rollup-plugin-babel'
 const pkg = require('./package.json')
+import css from 'rollup-plugin-css-porter'
 
 export default {
   entry: 'src/index.js',
   dest: pkg.main,
   format: 'cjs',
   sourceMap: true,
-  plugins: babel()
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -1,10 +1,11 @@
...
 const pkg = require('./package.json')
+import css from 'rollup-plugin-css-porter'
 
 export default {
   entry: 'src/index.js',
   dest: pkg.main,
   format: 'cjs',
   sourceMap: true,
-  plugins: babel()
+  plugins: [ css({ minified: false }), babel() ]
```
niedzielski: Ah! I think the reason you had said that we weren't doing Less was because this plugin doesn't handle it?

montehurd: Reminder: Update the changelog with something noting this css ability.

PR #7, opened 2017-03-22
Ignore MacOS '.DS_Store' files.

montehurd:
MacOS adds these everywhere :/

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,4 @@
 /build/
+
+# Mac OS X ignores
+.DS_Store


niedzielski:
> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig.

I think better practice is to place the exclude in a global Git [excludesfile](https://git-scm.com/docs/gitignore/2.9.1) per user. It's obviously a petty thing to not merge so let me know what you'd prefer!

montehurd:
Oh thanks that makes sense!

montehurd:
Closing per @niedzielski's comments.

PR #8, opened 2017-03-24
1st pass at consolidation of common image widening bits.

montehurd:


montehurd:
Integration of this on iOS: https://github.com/wikimedia/wikipedia-ios/pull/1313

test/Utilities/FixtureIO.js
```diff
@@ -0,0 +1,16 @@
+import domino from 'domino'
+const fs = require('fs')
```
niedzielski: Any reason not to use imports? e.g.:

```js
import fs from 'fs'
import path from 'path'
```

src/WidenImage.js
```diff
@@ -0,0 +1,86 @@
+import './WidenImage.css'
+import utilities from './Utilities'
+
+// If enabled, widened images will have thin red dashed border
+const enableDebugBorders = false
```
niedzielski: If the intent is to allow runtime debugger changes to this value, do we need this to be `let` (with an `// eslint-disable-line prefer-const`) instead of `const`? In Chromium, when I type `const foo = 0; foo = 1` it throws a `TypeError`.

test/WidenImage.js
```diff
@@ -0,0 +1,78 @@
+import applib from '../build/applib'
+import assert from 'assert'
+import fixtureIO from './Utilities/FixtureIO'
+import styleMocking from './Utilities/StyleMocking'
+
+const maybeWidenImage = applib.WidenImage.maybeWidenImage
+let document = null
```
niedzielski: For whatever reason, I never thought to use a top level variable in my PR. Any case for or against 1) top level variable vs 2) `this.document` in `beforeEach()`? I'm not sitting on anything informative but we should be consistent one way or another.

.eslintrc.yml
```diff
@@ -13,4 +13,5 @@ rules:
   func-style: [2, expression]
   indent: [1, 2]
   quotes: [1, single, {avoidEscape: true}]
-  semi: [1, never]
\ No newline at end of file
+  semi: [1, never]
+  # no-console: 0
```
montehurd: Opps I'll get rid of that.

src/Utilities.js
```diff
@@ -0,0 +1,22 @@
+
```
montehurd: Good catch!

src/Utilities.js
```diff
@@ -0,0 +1,22 @@
...
+// Implementation of https://developer.mozilla.org/en-US/docs/Web/API/Element/closest
+const findClosest = (el, selector) => {
```
montehurd: Doh! Totally agree. Oh, I wonder if there's a linting option to enforce...

src/Utilities.js
```diff
@@ -0,0 +1,22 @@
...
+  while ((el = el.parentElement) && !el.matches(selector)) {
```
montehurd: This one's a bit weird because the while bails if either the assignment evaluation (iirc the first inner parenthesis around `el = el.parentElement` are required) or the `matches` attempt fails. I'll add some tests for this method so we can more freely fiddle with it :) Totally open to `no-cond-assign` set to "always".

src/Utilities.js
```diff
@@ -0,0 +1,22 @@
...
+    // Intentionally empty.
+  }
+  return el
+}
+
+const isNestedInTable = (el) => {
+  while ((el = el.parentElement)) {
```
montehurd: As mentioned in the comment above iirc these parenthesis actually are required to evaluate the assignment. Will double-check and add tests for this method too.

src/Utilities.js
```diff
@@ -0,0 +1,22 @@
...
+    if (el.tagName === 'TD') {
+      return true
+    }
+  }
+  return false
+}
+
+export default {
```
montehurd: Absolutely. Good idea.

src/WidenImage.css
```diff
@@ -0,0 +1,61 @@
+
```
montehurd: I made an attempt to do so after our last chat at the onsite/offsite but ran into a few complications. Then I did an audit on the iOS transform less files and didn't see any actual less syntax, so I was thinking we could go with plain css for now and re-evaluate later as needed if that sounded ok?

montehurd: I'm open to that. 

The only consideration I'd want to chat about is the extent to which we want to encourage modularity VS sharing - between transforms either in JS or CSS.

By having JS and accompanying CSS in same folder makes it maybe a little easier to think of them in pair-wise fashion (each JS has CSS file of same name if needed) rather than thinking of CSS folder as a place you put CSS to be shared between JS transform files.

...though we could outline best practice in readme.md once we decide what those are :)

src/WidenImage.css
```diff
@@ -0,0 +1,61 @@
...
+.wideImageOverride {
```
montehurd: Ohh thanks for the BEM link... I'll try to read up on that tomorrow.

As far as naming conventions go I'm pretty open to suggestions. Good chat topic for tomorrow.

src/WidenImage.css
```diff
@@ -0,0 +1,61 @@
...
+    /* Center images. */
+    display: block;
+    margin-left: auto;
+    margin-right: auto;
+}
+
+@media (orientation: portrait) {
+    .wideImageOverride {
+        width:100% !important;
```
montehurd: Sure!

src/WidenImage.css
```diff
@@ -0,0 +1,61 @@
...
+        max-width:100% !important;
+        height:auto !important;
```
montehurd: Ah good point. If the transform works without it I'd think removing it would be best.

src/WidenImage.css
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /* Tablet override */
+    @media (min-device-width: 768px) {
```
montehurd: I think I'd be down with that - though probably a bit less so _no pun intended_ if that were the only LESS feature we used.

src/WidenImage.css
```diff
@@ -0,0 +1,61 @@
...
+
+/* Wrangle extra margin - see enwiki "Claude Monet > Impressionism" */
+li.gallerybox div.thumb > div {
+    margin:0px !important;
+}
+div.gallerytext > p {
+    margin-top:0px !important;
+}
+
+/* Ensure widening can take effect with proportional height change */
```
montehurd: Ah that's a good point - although this does need to happen for widened images, it probably shouldn't be fiddled with for images which are not widened. I'll try moving it to `.wideImageOverride`.

src/WidenImage.js
montehurd: I'd only ever changed this manually, but honestly I think we should delete it. I had used it to make it _very_ apparent when an image had been widened by having it use loud colored borders, but I can just manually set such borders in `.wideImageOverride` when needed.

src/WidenImage.js
```diff
@@ -0,0 +1,86 @@
...
+
+const widenAncestors = (el) => {
+  while ((el = el.parentElement) && !el.classList.contains('content_block')) {
```
montehurd: Not strictly, but I bailed at the content block since images are ultimately inside content blocks - no need to keep walking up the DOM past that point.

src/WidenImage.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+}
+
+const shouldWidenImage = (image) => {
+  // Some images are within a <div class='noresize'>...</div> which indicates
+  // they should not be widened. Example below has links overlaying such an image.
+  // See:
+  //   'enwiki > Counties of England > Scope and structure > Local government'
+  if (utilities.findClosest(image, "[class*='noresize']")) {
```
montehurd: +1 to renaming `utilities` to `elementUtilities`. Narrow scope FTW!

iirc with `findClosest` I was trying to imitate [JQuery](https://www.w3schools.com/jquery/traversing_closest.asp) and [Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) which just use `closest`.

But point taken on names giving context... `closestAncestorMatchingSelector` is a bit verbose but pretty clear. Do you like that less/more than `queryClosestSelector`?

src/WidenImage.js
```diff
@@ -0,0 +1,86 @@
...
+  if (enableDebugBorders) {
+    image.classList.add('wideImageDebug')
+  }
+}
+
+const maybeWidenImage = (image) => {
+  if (shouldWidenImage(image)) {
+    widenImage(image)
+    return true
+  } else {
```
montehurd: Hmm but isn't that how the caller knows when an image wasn't widened?

test/Utilities/FixtureIO.js
```diff
@@ -0,0 +1,16 @@
+import domino from 'domino'
```
montehurd: Works for me! Agreed - good readme addition!

test/Utilities/FixtureIO.js
montehurd: Oh good catch I'll try it.

test/Utilities/StyleMocking.js
```diff
@@ -0,0 +1,40 @@
...
+const mockStylesInElement = (element, styles) => {
+  for (const style in styles) {
+    if ({}.hasOwnProperty.call(styles, style)) {
+      element.style[style] = styles[style]
+    }
+  }
+}
+
+const verifyStylesInElement = (element, styles) => {
+  for (const style in styles) {
```
montehurd: I don't recall any reason not to use `for..of` - will try it.

test/Utilities/StyleMocking.js
```diff
@@ -0,0 +1,40 @@
+import assert from 'assert'
+
+// Domino doesn't seem to slurp up css class properties if we set them in our fixture html.
```
montehurd: Hmm not sure I follow... the comment describes (perhaps poorly haha) how I'm using this file for testing, but perhaps proper method comments would be better?

test/Utilities/StyleMocking.js
```diff
@@ -0,0 +1,40 @@
...
+  for (const style in styles) {
+    if ({}.hasOwnProperty.call(styles, style)) {
+      const msg = `'${style}' should be '${styles[style]}' but was '${element.style[style]}'`
+      assert.ok(element.style[style] === styles[style], msg)
+    }
+  }
+}
+
+const mockStylesInElements = (elements, styles) => {
+  for (let i = 0; i < elements.length; i++) {
```
montehurd: Totally agree :) I'll add JSDocs all around.

montehurd: And I'll switch to `for (const el of elements)`.

test/WidenImage.js
montehurd: I think I'm pretty against top level vars if they can be avoided. Perhaps less concerned in test files like this, but your `this.document` in `beforeEach()` suggestion should achieve same effect - I'll try switching to that.

test/WidenImage.js
```diff
@@ -0,0 +1,78 @@
...
+  beforeEach(() => {
+    document = fixtureIO.documentFromFixtureFile('WidenImage.html')
+  })
+
+  describe('Images which should not widen', () => {
+    it('image has usemap', () => {
+      const image = document.getElementById('imageWithUsemap')
+      const result = maybeWidenImage(image)
+      assert.ok(result === false)
+      assert.ok(image.classList.contains('wideImageOverride') === false)
```
montehurd: Oh that's more concise... haha wonder why I didn't do that :)

test/WidenImage.js
```diff
@@ -0,0 +1,78 @@
...
+
+      styleMocking.mockStylesInElements(ancestors, {
+        width: '50%',
+        maxWidth: '50%',
+        float: 'right'
+      })
+
+      const image = document.getElementById('imageInWidthConstrainedAncestors')
+      const result = maybeWidenImage(image)
+      assert.ok(result === true)
```
montehurd: Hmm I wonder if linting complained or something... I'll check/switch.

test/fixtures/WidenImage.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
montehurd: Interesting! Ya I can switch that. Was using...

    const ancestors = document.querySelectorAll("[id*='widthConstrainedAncestor']")

... but can just as easily use a class.

montehurd: Ah, can do!

src/WidenImage.js
```diff
@@ -0,0 +1,86 @@
...
+    if (el.style.maxWidth) {
+      el.style.maxWidth = '100%'
+    }
+    if (el.style.float) {
+      el.style.float = 'none'
+    }
+  }
+}
+
+const shouldWidenImage = (image) => {
```
montehurd: I think I'm fine with that, though in general I tend to default to private unless prodded :)

src/WidenImage.css
niedzielski: :+1: sounds good! BTW, I believe this CSS uncovers a [bug in the CSS validator](https://github.com/w3c/css-validator/issues/72).

niedzielski: :+1: Thanks for your thoughts on this. Let's keep on thinking but it's nothing to worry about now.

src/WidenImage.css
niedzielski: I don't mind if we _pun_t on this for now but if we do, would you mind adding our definition of wide screen width to the readme?

FWIW, it looks like CSS does actually have variable support though:

- https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables
- http://caniuse.com/css-variables/embed

I've unknowingly used these in the tiny bit of Polymer I tried but I thought it was like a hack specific to web components :D

src/WidenImage.js
niedzielski: I'm down with whatever. Less code is often good though!

src/WidenImage.js
niedzielski: Let's chat about this in the meeting then. My position is that in addition to providing clients a useful read-only method (doesn't mutate), it allows you to split the unit tests so they only have to do one assertion per case.

src/WidenImage.js
niedzielski: ..I didn't know about `closest()`. It sounds much _closer_ to the functionality provided so +1 for keeping the word "closest" in the name. I don't really mind if you do or don't tack other words on.

src/WidenImage.js
niedzielski: Sorry, I wasn't clear. I meant, replace:

```js
} else {
  return false
}
```

With:

```js
}
return false
```

test/Utilities/StyleMocking.js
niedzielski: Let's chat about this in the meeting.

src/WidenImage.js
niedzielski: From the meeting: will at least give a try to exporting a _shouldWidenImage and update the tests to be single assertion.

test/Utilities/StyleMocking.js
niedzielski: This was a profound discussion with a lot of subtelties uncovered. The minutes I took for our readme (both our responsibilities) are:

- Always prefer to modify the class list instead of inline style attributes and to test accordingly
- Always prefer to check the class list and inline styles instead of computed styles
- Add integration tests for parsoid output -- thehse are indepedent of styles
- Add visual regression tests for verifying cascading styles look good

^^^ FOr these rules above, we need some nice exampels for these above. And, just in general, we some nice examples. ^^^

.eslintrc.yml
montehurd: Fixed!

src/Utilities.js
montehurd: Fixed!

src/WidenImage.js
montehurd: I renamed `Utilities.js` to `ElementUtilities.js`. Pondering the function rename...

test/fixtures/WidenImage.html
montehurd: Fixed!

montehurd: Fixed!

src/WidenImage.css
montehurd: Fixed!

montehurd:
@niedzielski I got about half of the feedback issues fixed. Will try to knock out the remaining bits tomorrow. Oh also pushed up some [readme updates](https://github.com/wikimedia/applib/pull/9).

src/WidenImage.js
montehurd: Hey quick question - any idea why this doesn't work?

    export default {
        _shouldWidenImage = shouldWidenImage,
        maybeWidenImage
    }

montehurd: Fixed!

src/WidenImage.js
montehurd: Fixed!

montehurd: Oh, I'm onboard with adding `no-else-return` to the lint config.

test/WidenImage.js
montehurd: Fixed!

test/WidenImage.js
montehurd: Fixed!

test/Utilities/FixtureIO.js
montehurd: Fixed!

test/Utilities/FixtureIO.js
montehurd: Fixed!

src/WidenImage.js
montehurd: Fixed!

test/Utilities/StyleMocking.js
montehurd: Fixed!

test/Utilities/StyleMocking.js
montehurd: I switched to `for..of`.

src/Utilities.js
montehurd: Tests added! :)

montehurd: Confirmed the parenthesis are needed.

src/Utilities.js
montehurd: Confirmed the parenthesis are needed.

src/Utilities.js
montehurd: Fixed the table header bug and added basic tests.

src/WidenImage.css
montehurd: Moving `height: auto` to `.wideImageOverride` worked nicely.

src/WidenImage.js
montehurd: Opps I think I misread your comment - for some reason I wasn't thinking `unit` test. I just added one.

montehurd:
@niedzielski I think I've attempted to fix all feedback issues with the exception of adding JSDoc's all around, which I'll tackle next :)

test/WidenImage.js
niedzielski: I think this worked ok for me here: https://github.com/wikimedia/applib/pull/3/files#diff-c00bff523716354efebd88904a328d70R71

test/utilities/FixtureIO.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {?string}            String from `test/fixtures/${fileName}`
+ */
+const stringFromFixtureFile = (fileName) => {
+  return fs.readFileSync(path.resolve(__dirname, `../fixtures/${fileName}`), 'utf8')
+}
+
+/**
+ * Gets Domino document from file in 'test/fixtures/'
+ * @param  {!string} fileName   Name of file in 'test/fixtures/'
+ * @return {?document}          Domino document from `test/fixtures/${fileName}`
```
niedzielski: I believe this should be `Document` with a capital D: https://developer.mozilla.org/en-US/docs/Web/API/Document

src/ElementUtilities.js
```diff
@@ -0,0 +1,30 @@
...
+    // Intentionally empty.
+    // Reminder: the parenthesis around 'el = el.parentElement' are also intentional.
+  }
+  return el
+}
+
+/**
+ * Determines if element has a table ancestor.
+ * @param  {!HTMLElement}  el   Element
+ * @return {Boolean}            Whether table ancestor of 'el' is found
```
niedzielski: Consider using `boolean` with a lowercase b to match your lowercase `string` primitive typing

test/utilities/StyleMocking.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+const mockStylesInElements = (elements, styles) => {
+  for (const element of elements) {
+    mockStylesInElement(element, styles)
+  }
+}
+
+/**
+ * Verifies styles on each element in an array of elements.
+ * @param  {!Array<HTMLElement>}  elements  Array of elements
```
niedzielski: I read recently that there's a cleaner syntax for this `HTMLElement[]`.

http://usejsdoc.org/tags-type.html

test/utilities/FixtureIO.js
montehurd: Fixed!

src/ElementUtilities.js
montehurd: Fixed!

test/utilities/StyleMocking.js
montehurd: Oh nice! Fixed!

montehurd: I was using that page too but missed the array shorthand! :)

src/ElementUtilities.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Returns closest ancestor of element which matches selector.
+ * Similar to 'closest' methods as seen here:
+ *  https://api.jquery.com/closest/
+ *  https://developer.mozilla.org/en-US/docs/Web/API/Element/closest
+ * @param  {!HTMLElement} el    Element
```
niedzielski: Feel free to ignore this: this function appears to only actually only require [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element), not [HTMLElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement) specifically. For our code, I think the only time we'll generally require an HTMLElement is when using HTMLElement.style or HTMLElement.click() and then prefer Element or [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node).

montehurd: Fixed!

test/utilities/StyleMocking.js
```diff
@@ -0,0 +1,62 @@
+import assert from 'assert'
+
+// These methods are useful in tests which verify changes a method makes to element 'style'
+// properties.
+
+/**
+ * An object whose properties are taken to be style property names,
+ * and whose property values are taken to be style property values.
+ * @typedef {Object} StylesHash
```
niedzielski: This syntax is neat (and new to me)! It would be even cooler if we could represent a map. I looked in the docs but it wasn't quite clear to me.

http://usejsdoc.org/tags-type.html
http://usejsdoc.org/tags-typedef.html

montehurd: Something like... `{Object.<string, (string|number)>}`

niedzielski: That at least gives typing. I was looking at the `@prop` syntax but the trouble is that the property name varies.

niedzielski: Let's keep our eyes peeled when looking at other repos and maybe we'll stumble upon it one day!

niedzielski:
The bar is set!

PR #13, opened 2017-04-11
Chore: make CollapseTable tests more vebose

niedzielski:


niedzielski:
@montehurd, here's that patch to make the test more verbose and consistent with WidenImages that I promised. Let me know what you think of this style. I'm happy to rework this or abandon it.

PR #15, opened 2017-04-11
Chore: improve NPM package scripts

niedzielski:
Improve NPM package scripts:

- Use HEAD instead of master since origin/master and master may be
  divergent.
- Only lint when committing or publishing since style lints are fine
  during development.
- Update description and keywords to mention JavaScript and CSS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "applib",
   "version": "1.2.0",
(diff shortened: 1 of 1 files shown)


niedzielski:
@montehurd, "only lint when comitting or publishing" per https://github.com/wikimedia/applib/pull/8#discussion_r109513517.

PR #18, opened 2017-04-17
Chore: use consistent spacing in WidenImage CSS and JS

niedzielski:
The rest of the repo uses two space indentation. Update WidenImage.css
to use two spaces instead of four.


niedzielski:
Merging per earlier discussion and approval.

PR #21, opened 2017-04-19
Chore: favor undefined to null

niedzielski:
https://basarat.gitbooks.io/typescript/docs/tips/null.html
https://github.com/eslint/eslint/issues/6701


src/ElementUtilities.js
```diff
@@ -21,7 +21,7 @@ const findClosest = (el, selector) => {
  * @return {boolean}        Whether table ancestor of 'el' is found
  */
 const isNestedInTable = (el) => {
-  return (findClosest(el, 'table') !== null)
+  return (findClosest(el, 'table'))
```
niedzielski: Use `return !!findClosest(el, 'table')` to be explicit

src/ElementUtilities.js
```diff
@@ -21,7 +21,7 @@ const findClosest = (el, selector) => {
  * @return {boolean}        Whether table ancestor of 'el' is found
```
niedzielski: Please also update findClosest and other methods to eliminate parentheses usage.

niedzielski: This probably deviates from the original implementation in that it only considers the parent, not the passed in element. Consider replacing with the original and replacing usages with el.parentElement OR renaming to findClosestAncestor().

niedzielski: Also, regardless: add a test for the case of whether the element itself can / is expected to be returned when it matches.

niedzielski: On second thought, just go with renaming and adding that test and updating the assignment / parentheses usage.

niedzielski: Do the same thing for other methods elsewhere including WidenImage.widenAncestors().

niedzielski: Please put in separate PR yo.

src/ElementUtilities.js
niedzielski: Fixed!

src/ElementUtilities.js
niedzielski: Ok, this is all done but it landed in this PR :]

montehurd:
I'm a bit worried about the readability trade-offs now that I see the diff... the for loops and `!!` are less clear at first glance, but I'm not sure if that's just because I'm used to the old way... I'll take another peek tomorrow when I'm not sleepy :)

PR #22, opened 2017-04-19
Breaking: rename applib to wikimedia-page-library

niedzielski:
This patch renames applib to wikimedia-page-library. As part of this
change, breaking build product renames are made for the JavaScript and
CSS outputs and the NPM references are updated. `pagelib` is used in
JavaScript to keep code at a readable balance between bloat and cryptic.

### DIFF
--- a/test/CollapseTable.js
+++ b/test/CollapseTable.js
@@ -1,10 +1,10 @@
-import applib from '../build/applib'
 import assert from 'assert'
 import domino from 'domino'
+import pagelib from '../build/wikimedia-page-library'
 
 describe('CollapseTable', () => {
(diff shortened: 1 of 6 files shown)


test/CollapseTable.js
```diff
@@ -1,10 +1,10 @@
-import applib from '../build/applib'
 import assert from 'assert'
 import domino from 'domino'
+import pagelib from '../build/wikimedia-page-library'
```
niedzielski: Let me know if `pagelib` works for you! `wikimediaPageLibrary.module.method` felt too verbose for code.

niedzielski:
Merging per earlier and approval.

PR #23, opened 2017-04-19
Breaking: divide outputs into transform and overrides

niedzielski:


niedzielski:
Merging per earlier discussion and approval.

PR #25, opened 2017-05-04
Introduce polyfill for matching element with selector.

dbrant:
Some older versions of the Android WebView do not provide the `matches()`
method on its elements. They do, however, provide an equivalent method
called `webkitMatchesSelector()`, and I've also observed a Samsung flavor
of WebView that provides `matchesSelector()`. This patch introduces a
polyfill that tries to use all three of these variations.

This is currently causing a regression where image widening is not working in API <19.

### DIFF
--- a/src/transform/ElementUtilities.js
+++ b/src/transform/ElementUtilities.js
@@ -2,7 +2,7 @@
  * Polyfill function that tells whether a given element matches a selector.
  * @param {!Element} el Element
  * @param {!string} selector Selector to look for
- * @returns {boolean} Whether the element matches the selector
+ * @returns {!boolean} Whether the element matches the selector
  */
 const matchesSelectorCompat = (el, selector) => {
   if (el.matches) {
--- a/src/transform/ElementUtilities.js
+++ b/src/transform/ElementUtilities.js
@@ -1,4 +1,21 @@
 /**
+ * Polyfill function that tells whether a given element matches a selector.
+ * @param {!Element} el Element
+ * @param {!string} selector Selector to look for
+ * @returns {boolean} Whether the element matches the selector
+ */
+const matchesSelectorCompat = (el, selector) => {
+  if (el.matches) {
+    return el.matches(selector)
+  } else if (el.matchesSelector) {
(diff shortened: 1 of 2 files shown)


src/transform/ElementUtilities.js
```diff
@@ -1,4 +1,21 @@
 /**
+ * Polyfill function that tells whether a given element matches a selector.
+ * @param {!Element} el Element
+ * @param {!string} selector Selector to look for
+ * @returns {boolean} Whether the element matches the selector
+ */
+const matchesSelectorCompat = (el, selector) => {
```
montehurd: Oh this is great! 

The only thing I'm wondering about is the naming... does `safeMatches` sound any better/worse?

dbrant: Sure! I used the `...Compat` nomenclature reflexively because it's very common in the Android ecosystem, but `safeMatches` does sound more friendly.

niedzielski: @mhurd, `compat` is an Android convention. Do you mind if I take an action item for this? I was hoping we could discuss polyfill strategy in general.

montehurd: @niedzielski Sounds good!

montehurd: Could we just go with `matchesCompat` for now?

montehurd: (I don't feel strongly about it - so can totally wait too.)

niedzielski:
I've taken the following action items:
- Bump the version.
- Add sync meeting item to discuss how to handle polyfills in wikimedia-page-library. This implementation is very focused. The [MDN implementation](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches) updates the prototype chain. If still applicable, discuss naming convention.
- Add tests.
- Add no unnecessary else lint checkers.

@mhurd, since you've approved I'm merging this to unblock @dbrant but agree we need to discuss the second bullet further.

niedzielski:
@dbrant published in v2.0.1. Thanks for the cool fix!

PR #26, opened 2017-05-04
Update .gitignore

dbrant:
Adds a few rules to ignore IDE files and OS-specific stuff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 /node_modules/
 /.eslintcache
 /build/
+
+.DS_Store
+.idea
+*.iml


.gitignore
```diff
@@ -1,3 +1,7 @@
 /node_modules/
 /.eslintcache
-/build/
\ No newline at end of file
+/build/
+
+.DS_Store
+.idea
```
montehurd: lol no ideas allowed :)

.gitignore
```diff
@@ -1,3 +1,7 @@
 /node_modules/
 /.eslintcache
-/build/
\ No newline at end of file
+/build/
+
+.DS_Store
```
niedzielski: > Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig.

I think it would be better practice is to place the exclude in a global Git [excludesfile](https://git-scm.com/docs/gitignore/2.9.1) per user. I don't mind us versioning project files for editors that are in use (for example, if you have WebStorm project file) but is there a case where any repo would ever not want to ignore a .DS_Store / .iml / .idea file? For instance, I could see this file growing to include Vim backup files, Windows' Thumbs.db, etc and it kind of feels out of scope. Thoughts?

montehurd: oh this rings a bell - you had me do something similar a while back - adding such things to my global user ignore file

niedzielski:
Thanks for the due diligence but we're closing this out after discussing. Please reopen if you disagree!

PR #29, opened 2017-05-04
Chore: add tests for ElementUtilties.matchesSelectorCompat()

niedzielski:


src/transform/ElementUtilities.js
```diff
@@ -44,6 +46,7 @@ const isNestedInTable = (el) => {
 }
 
 export default {
+  matchesSelectorCompat,
```
montehurd: Oh could we use a `test` container here?
<pre>
export default {
  findClosestAncestor,
  isNestedInTable,
  test: {
    matchesSelectorCompat
  }
}
</pre>

niedzielski: I can ElementUtilities is already under test though. This would end up looking like `pagelib.test.ElementUtilities.test.matchesSelectorCompat`.

PR #30, opened 2017-05-08
Chore: omit needless parentheses and braces on arrow functions

niedzielski:
Update the ESLint overrides to require parentheses ("()") and curly
braces ("{}") on arrow functions only as needed.


niedzielski:
@montehurd, per meeting minutes, if I took them down right!

PR #43, opened 2017-06-28
Breaking: emit a `section-toggled` event

niedzielski:
- Breaking: require a Window instead of a Document in the first
  parameter of CollapseTable.collapseTables().
- New: emit a `section-toggled` CustomEvent when a table is expanded or
  collapsed. This is needed by lazily loaded images to monitor layout
  changes.


niedzielski:
@montehurd, please review when you can. This is needed by the lazily loaded images implementation.

montehurd:
Hmm I'm a little foggy on why the window is needed...

PR #46, opened 2017-07-03
Fix: VS Code launch config

niedzielski:
- Force new protocol.
- Fix test glob.

https://code.visualstudio.com/updates/v1_10#_node2-transitioning
https://github.com/Microsoft/vscode/issues/21883


montehurd:
@niedzielski gonna trust you on this :) I don't use VS.

PR #48, opened 2017-07-05
Use `Array.from` to fix Safari/iOS 9 bug.

montehurd:
https://phabricator.wikimedia.org/T169660

### DIFF
--- a/src/transform/RedLinks.js
+++ b/src/transform/RedLinks.js
@@ -12,9 +12,10 @@ const configureRedLinkTemplate = (span, anchor) => {
 /**
  * Finds red links in a document or document fragment.
  * @param {!(Document|DocumentFragment)} content Document or fragment in which to seek red links.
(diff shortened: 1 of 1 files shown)


src/transform/RedLinks.js
```diff
@@ -12,9 +12,10 @@ const configureRedLinkTemplate = (span, anchor) => {
 /**
  * Finds red links in a document or document fragment.
  * @param {!(Document|DocumentFragment)} content Document or fragment in which to seek red links.
- * @return {!NodeList} Nodelist of zero or more red link anchors.
+ * @return {!HTMLAnchorElement[]} Array of zero or more red link anchors.
  */
-const redLinkAnchorsInContent = content => content.querySelectorAll('a.new')
+const redLinkAnchorsInContent =
+  content => Array.prototype.slice.call(content.querySelectorAll('a.new'))
```
montehurd: Note: we didn't use `Array.from` because it wasn't supported by all Android targets.

PR #52, opened 2017-07-09
Doc: update changelog

niedzielski:


niedzielski:
I'd like to cut a release with just this upgrade. If any problems arise during upcoming feature integration (lazy loading, footer) this should hopefully make it simpler to attribute or disattribute to the upgrades.

PR #53, opened 2017-07-12
New: Footer transforms

montehurd:
https://phabricator.wikimedia.org/T164137

As these are consolidated I'm concurrently updating [this iOS PR](https://github.com/wikimedia/wikipedia-ios/pull/1615).

New transforms:
- [x] container
- [x] menu
- [x] read more
- [x] legal

Misc:
- [x] compatibility for older devices (audit array methods etc)

----

Needed follow-ups:
- expand test coverage
- add demos
- consider further internals refactors/simplifications. 

Feedback comments which I've yet to address or which we haven't come to consensus on yet:
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50064241
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50504536
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50502046
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50236721
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50235927
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50235552
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50232915
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50232674
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50087980
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50083151
https://github.com/wikimedia/wikimedia-page-library/pull/53#pullrequestreview-50068612


montehurd:
Wondering if we should enable the [valid-jsdoc](http://eslint.org/docs/rules/valid-jsdoc) eslint rule...

niedzielski:
@montehurd, yes! Please enable valid-jsdoc. I'm sorry but I thought we already had it enabled and was mistaken.

src/transform/FooterContainer.css
```diff
@@ -0,0 +1,25 @@
+
+.footer_container {
```
niedzielski: It's fine to punt on this but consider using a common prefix like pagelib-footer-thing.

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,85 @@
...
+ * implementation details.
+ * @param {!Document} document
+ * @param {!number} margin
+ * @return {void}
+ */
+const updateLeftAndRightMargin = (document, margin) => {
+  const elements = document.querySelectorAll(
+    '#footer_container_menu_heading, #footer_container_readmore, #footer_container_legal'
+  )
+  Array.from(elements)
```
niedzielski: I see you already called this out in the todo but consider replacing `Array.from()` with `Array.prototype.slice.call()` for old time Android devices.

src/transform/FooterMenu.css
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.footer_menu_icon_disambiguation {
+    background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAFAAAABQCAYAAACOEfKtAAAAAXNSR0IArs4c6QAAAWRJREFUeAHt1ruOwjAQBVAW8YnbbMHXUdDwj6xWKyigGJLrcaTo0PBwBifH1xofDl4ECBAgQIAAAQIECBAgQIAAAQIECBAgQGCGwNeMSf7m+P4532fNtWSe2/USGRyXTObad4HT+0+9v6QrPuruRu0ICQxXBCDAUCAsl0CAoUBYPqwLf9rVPr0ufK5neXfXt4Wf1Os+DEvgY/ruFX/MU73PSroEVitRjAMsgKphgJVQMQ6wAKqGAVZCxTjAAqgaBlgJFePDz4Gv8806j211/pTA1xVf+L09gVslY6HD6sslcDXdfyFAgKFAWC6BAEOBsLy9CzsHhiu09/L2BDoH7j1C4fPpwgBDgbBcAgGGAmF5exd2DgxXaO/l7Qnc+hzYvQM0kXCLtCcwvL+4vHsHSGC4RAABhgJhuQQCDAXC8uFduPvcFT7v8HJbeDipPyRAgAABAgQIECBAgAABAgQIECBAYK8Cv3qKKmmwPBETAAAAAElFTkSuQmCC);
```
niedzielski: I wonder why GitHub is coloring this line differently.

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,215 @@
...
+  * @typedef {function} FooterMenuItemClickCallback
+  * @param  {!string[]} payload Important - should return empty array if no payload strings.
+  * @return {void}
+  */
+
+/**
+ * Extracts array of no-html page issues strings from document.
+ * @type {FooterMenuItemPayloadExtractor}
+ */
+const pageIssuesStringsArray = document => {
```
niedzielski: This seems to be a method but uses `@type`.

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /**
+   * Returns reference to function for extracting payload when this menu item is tapped.
+   * @return {FooterMenuItemPayloadExtractor}
+   */
+  payloadExtractor() {
+    switch (this.itemType) {
+    case MenuItemType.languages:
+      return null
```
niedzielski: Please use `undefined` instead of `null` per earlier discussion.

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,215 @@
...
+      subtitle.innerText = wmfMenuItem.subtitle
+      containerAnchor.appendChild(subtitle)
+    }
+
+    const iconClass = wmfMenuItem.iconClass()
+    if (iconClass) {
+      item.classList.add(iconClass)
+    }
+
+    return document.createDocumentFragment().appendChild(item)
```
niedzielski: Is it intentional to a return a value from the constructor?

src/transform/index.js
```diff
@@ -9,6 +12,9 @@ import WidenImage from './WidenImage'
 
 export default {
   CollapseTable,
+  FooterContainer,
```
niedzielski: This is great how you have these broken out!

src/transform/FooterLegal.css
```diff
@@ -0,0 +1,38 @@
...
+
+html[dir="ltr"] .footer_legal_contents {
+  background-position: left top;
+}
+
+html[dir="rtl"] .footer_legal_contents {
+  background-position: right top;
+}
+
+.footer_legal_licence {
```
montehurd: Typo!

src/transform/FooterContainer.css
montehurd: Good idea! Once I get `Read more` into this PR I'll make a wholesale change.

src/transform/FooterMenu.css
montehurd: Ha I was wondering about this too!

src/transform/FooterMenu.js
montehurd: See the `FooterMenuItemPayloadExtractor` typedef for the details. I think I did this correctly according to the `Type definitions` box in the [docs](http://usejsdoc.org/tags-type.html). Lemme know if not :)

src/transform/FooterContainer.css
montehurd: Fixed! (I didn't change from underscores to dashes yet though)

src/transform/FooterLegal.css
montehurd: Fixed.

src/transform/FooterContainer.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.pagelib_footer_container_heading {
+    padding-top: 35px;
+    font-size: 0.95em;
+    line-height: 1.6;
+}
+
+#pagelib_footer_container_readmore_pages::-webkit-scrollbar {
+  display: none;
```
niedzielski: I'm sure you tried this already but does `overflow: hidden` not work?

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,92 @@
+import './FooterContainer.css'
+
+/**
+ * Ensures the 'Read more' section header can always be scrolled to the top of the screen.
+ * @param {!Document} document
+ * @param {!Window} window
+ * @return {void}
+ */
+const updateBottomPaddingToAllowReadMoreToScrollToTop = (document, window) => {
+  const div = document.getElementById('pagelib_footer_container_ensure_can_scroll_to_top')
```
niedzielski: Consider making a symbol for this ID.

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,92 @@
+import './FooterContainer.css'
+
+/**
+ * Ensures the 'Read more' section header can always be scrolled to the top of the screen.
+ * @param {!Document} document
```
niedzielski: Do we want to have a document parameter or can we just use `window.document`?

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,92 @@
...
+ * @param {!Window} window
+ * @return {void}
+ */
+const updateBottomPaddingToAllowReadMoreToScrollToTop = (document, window) => {
+  const div = document.getElementById('pagelib_footer_container_ensure_can_scroll_to_top')
+  let currentPadding = parseInt(div.style.paddingBottom, 10)
+  if (isNaN(currentPadding)) { currentPadding = 0 }
```
niedzielski: Since NaN is falsy, I think the preceeding line could just be currentPadding = parseInt(...) || 0

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,92 @@
...
+  const height = div.clientHeight - currentPadding
```
niedzielski: Do we need to check border width and margin here too?

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,215 @@
...
+  const tables = document.querySelectorAll(
+    'div#content_block_0 table.ambox:not(.ambox-multiple_issues):not(.ambox-notice)'
+  )
+  // Get the tables into a fragment so we can remove some elements without triggering a layout
+  const fragment = document.createDocumentFragment()
+  for (let i = 0; i < tables.length; i++) {
+    fragment.appendChild(tables[i].cloneNode(true))
+  }
+  // Remove some element so their text doesn't appear when we use "innerText"
+  Array.from(fragment.querySelectorAll('.hide-when-compact, .collapsed')).forEach(el => el.remove())
```
niedzielski: Since I ended up with a lot of these in my own code, I ended up adding a polyfill in my patch for querySelectorAll.

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,215 @@
...
+ */
+const disambiguationTitlesArray = document => Array.from(
+  document.querySelectorAll(
+    'div#content_block_0 div.hatnote a[href]:not([href=""]):not([redlink="1"])'
+  )
+).map(el => el.href)
+
+/**
+ * Type representing kinds of menu items.
+ * @type {Object}
```
niedzielski: I'm still fuzzy on how these `@type` / `@typedef` are intended to work but I think this should be `@type {!MenuItemType}` and in another comment you'd do `@typedef {{name: !string, id: !number}} MenuItemType` or

```
@typedef {!object} MenuItemType
@prop {!number} languages
@prop {!number} lastEdited
```

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,215 @@
...
+    this.clickHandler = clickHandler
+    this.payload = []
+  }
+
+  /**
+   * Returns icon CSS class for this menu item based on its type.
+   * @return {!string}
+   */
+  iconClass() {
+    switch (this.itemType) {
```
niedzielski: An alternative to this switch might be:

```js
class MenuItemType {
  constructor(id, iconClass) {
    this._id = id
    this._iconClass = iconClass
  }
  get id() { return this._id }
  get iconClass() { return this._iconClass }
}

const MENU_ITEMS = {
  LANGUAGES: new MenuItemType(1, 'pagelib-footer-menu-icon-languages'),
  LAST_EDITED: new MenuItemType(2, 'pagelib-footer-menu-icon-last-edited'),
  PAGE_ISSUES: new MenuItemType(3, 'pagelib-footer-menu-icon-page-issues'),
  DISAMBIGUATION: new MenuItemType(4, 'pagelib-footer-menu-icon-disambiguation'),
  COORDINATE: new MenuItemType(5, 'pagelib-footer-menu-icon-coordinate')
}
```
Then the `WMFMenuItem.iconClass()` implementation becomes `return menuItem.iconClass()` and you don't have to remember to update it when adding new types.

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,215 @@
...
+  lastEdited: 2,
+  pageIssues: 3,
+  disambiguation: 4,
+  coordinate: 5
+}
+
+/**
+ * Menu item model.
+ */
+class WMFMenuItem {
```
niedzielski: Everything in the library will probably be WMF-ish. Any reason not to call this just "MenuItem"?

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,349 @@
...
+  )
+}
+
+/**
+ * Sets heading element string.
+ * @param {!string} headingString
+ * @param {!string} headingID
+ * @param {!Document} document
+ */
+const setHeading = (headingString, headingID, document) => {
```
niedzielski: Consider consistently placing the document / window parameter in the same position in each method signature unless there's a good reason to deviate.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,349 @@
...
+   * @typedef {function} ShownReadMoreHandler
+   * @param {!Object[]} pages
+   * @param {!Document} document
+   * @param {!string} containerID
+   * @param {SaveButtonClickHandler} saveButtonClickHandler
+   * @param {TitlesShownHandler} titlesShownHandler
+   * @return {void}
+   */
+
+let _saveForLaterString = null
```
niedzielski: These globals suggest that a class implementation may be more appropriate. The client can than keep one or more instances of that class as needed.

src/transform/FooterReadMore.css
```diff
@@ -0,0 +1,92 @@
...
+.pagelib_footer_readmore_bookmark_filled {
+    background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADAAAABGCAYAAACKeRdbAAAABGdBTUEAALGPC/xhBQAAAAFzUkdCAK7OHOkAAAAgY0hSTQAAeiYAAICEAAD6AAAAgOgAAHUwAADqYAAAOpgAABdwnLpRPAAAAAZiS0dEAP8A/wD/oL2nkwAAAqhJREFUaN7t20lrE3EYx/HfzGQyM5nsNmkTEo2JJmmSCW5RRHsIir20B0GKCuJFg6CHllKV1rbULBeh4EFcMCIEN2xLjEevHn0NBREEwfcQDzWFUJNMNp8E/s8tQ5L/9xMmsxyGO575VsUQD08dwADUAQxAHcAA1AEMQB3AANQBDEAdwADUAQxAHcAA1AEMQB3AANQBDEAdwADUAQxAHcAA1AEMQB3AANQBDEAdwADUAQxAHcAA1AEMQB0w0IDJk05cSDn7CjD064tn0m7cvewHAJgkHuWvv4cHkJn2IjPl2X394NoBqIqAN19+DT5gfsaPK+fce7bPXfLBLAt4/vnngAI4YPX6QUyfbrzP35zyQJUFrH/8MXiAR7dCSB+xt3zf1fNuKDKPfOl7T9bt+ijEccCT2bCu+NpcPDuC/I0gOGoAzwHFhShOjVva/uxkyoH126GdX4ACIPBAaSmGZEjtePGJpB3PZg93ZegIYDTw+LAaR8SvdL7y3zkRtaC4EAXfIaJtgGzksZlNIDAmdx1fm2RIRWkpBoPQvqItgCoLKOc0eJxiz+JrE/EreL8Sg2hoD6EbYDUJqBQ0jNj6dvWBwJiMzYcJSKJ+hC6A0yqiUtBgU4W+xdfGu8+Ick6DSda3VkvAqEPEp1wCZqX/8bVx2UVU8glYTa3XbArwuSRsZTUo0v+/bbCbDagUNDgszf9vDcuCHhkba/G29sdej1kRUMknMOpojPgnILLfhHcr8Y4Oa70eReKxldXgc0n6AMmgitLiOIQButmURA4ba3GEvHtPnHWZqagFr+51flbs5xgEDm+XY4gH6i9ddgFnNBuezoWpO5uOwAOv70dxLGyuB6SPOvD4ziHqPl3DccCL+QgmkvadDcvF7eqwzuLL7SpXrVaH+iGgP5QPEShTwCAqAAAAAElFTkSuQmCC);
```
niedzielski: Consider adding an RTL variant if these can be flipped:

```css
  float: left;
  -webkit-transform: scaleX(-1);
  transform: scaleX(-1);
```

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,349 @@
...
+const _saveButtonIDPrefix = 'readmore:save:'
+
+/**
+ * Removes parenthetical enclosures from string. 
+ * @param {!string} string
+ * @param {!string} opener
+ * @param {!string} closer
+ * @return {!string}
+ */
+const safelyRemoveEnclosures = (string, opener, closer) => {
```
niedzielski: I know you're still working on writing a couple tests. This would probably be more worthwhile to exercise than some of the other methods.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,349 @@
...
+  titlesShownHandler(shownTitles)
+}
+
+/**
+ * Makes 'Read more' query parameters object for a title.
+ * @param {!string} title
+ * @param {!number} count
+ * @return {!Object}
+ */
+const queryParameters = (title, count) => ({
```
niedzielski: This network request stuff might be worth breaking into a separate file, ReadMoreClient or something like that, so we can separate concerns between UI and networking easily.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,349 @@
...
+ */
+
+/**
+  * @typedef {function} TitlesShownHandler
+  * @param {!string[]} titles
+  * @return {void}
+  */
+
+/**
+   * Display fetched read more pages.
```
niedzielski: The indent on this line is too great.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,349 @@
...
+ * @param {TitlesShownHandler} titlesShownHandler
+ * @param {!Document} document
+ * @return {void}
+ */
+const fetchReadMore = (baseURL, title, showReadMoreHandler, containerID, saveButtonClickHandler,
+  titlesShownHandler, document) => {
+  const xhr = new XMLHttpRequest() // eslint-disable-line no-undef
+  xhr.open('GET', readMoreQueryURL(title, baseURL), true)
+  xhr.onload = function() {
+    if (xhr.readyState === 4) {
```
niedzielski: How about `XMLHttpRequest.DONE`?

src/transform/FooterContainer.css
montehurd: I removed this - was unused.

src/transform/FooterContainer.js
montehurd: Fixed!

src/transform/FooterContainer.js
montehurd: Fixed!

src/transform/FooterMenu.js
montehurd: I copied your polyfill method over and used it :)

src/transform/FooterReadMore.js
montehurd: Oh good catch! Fixed!

src/transform/FooterReadMore.js
montehurd: Oops I probably misspoke about this! Checking...

src/transform/FooterContainer.js
montehurd: Fixed with polyfill.

src/transform/FooterMenu.js
montehurd: Fixed!

src/transform/FooterMenu.js
montehurd: Fixed!

src/transform/FooterReadMore.css
montehurd: I checked and it doesn't look like any of the menu icons need to be mirrored when RTL.

src/transform/FooterReadMore.js
montehurd: Fixed!

src/transform/FooterMenu.js
montehurd: How about [this](https://github.com/wikimedia/wikimedia-page-library/pull/53/commits/842a28fd3bd20f36e34281ee4c3582991b584507)?

src/transform/FooterMenu.js
montehurd: I chewed on this for a bit... thinking I'd prefer to keep `MenuItemType` an enum. Also was curious as to the reason for opposition to switch... Thoughts?

src/transform/FooterMenu.js
montehurd: Fixed! (Fragment makers can just be functions.)

src/transform/FooterReadMore.js
montehurd: Good news! I didn't need those globals after all. We were already calling a method from native land to set the save button icon state - I modified (& renamed) it to be passed the text to use for the current state of the save button at the same time.

montehurd: So...fixed!

src/transform/FooterContainer.js
montehurd: Ok to hold off on this for now? The template complicates doing this... it's filled with such ids...

montehurd:
@niedzielski I addressed most of the feedback. Will tackle remaining bits next week.

src/transform/FooterReadMore.js
montehurd: Fixed!

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,91 @@
+import './FooterContainer.css'
+import Polyfill from './Polyfill'
+
+/**
+ * Ensures the 'Read more' section header can always be scrolled to the top of the screen.
+ * @param {!Window} window
+ * @return {void}
+ */
+const updateBottomPaddingToAllowReadMoreToScrollToTop = window => {
+  const div = window.document.getElementById('pagelib_footer_container_ensure_can_scroll_to_top')
```
niedzielski: Would you consider a symbol for this since it's used in a couple places?

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,91 @@
...
+  const currentPadding = parseInt(div.style.paddingBottom, 10) || 0
```
niedzielski: I think this expects all units in pixels. Consider adding a comment here.

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,91 @@
...
+    #pagelib_footer_container_readmore, 
+    #pagelib_footer_container_legal
+  `)
+  Array.from(elements)
+    .forEach(element => {
+      element.style.marginLeft = `${margin}px`
+      element.style.marginRight = `${margin}px`
+    })
+  const rightOrLeft = document.querySelector('html').dir === 'rtl' ? 'right' : 'left'
+  Array.from(Polyfill.querySelectorAll(document, '.pagelib_footer_menu_item'))
```
niedzielski: Since you're using the polyfill which always returns an array, I don't think you need the Array.from any more.

src/transform/FooterMenu.js
niedzielski: :+1: this looks good! My mistake!

src/transform/FooterMenu.js
niedzielski: The reason is the elimination of conditional logic and the potential to fall out of sync between the conditional and the declaration, adding new entries is one line instead of a couple, and everything that embodies a menu item's distinction from other items is in one place.

WRT the discussion on how payloadExtractor() would look after this change, you just tack on the ID getter on to the switch and cases (if it's not possible to also embed this conditional into MenuItemType itself):

```js
  payloadExtractor() {
    switch (this.itemType.id) {
      case MenuItemType.pageIssues.id: return pageIssuesStringsArray
      case MenuItemType.disambiguation.id: return disambiguationTitlesArray
      default: return undefined
    }
  }
```

src/transform/FooterReadMore.css
```diff
@@ -0,0 +1,92 @@
+
+.pagelib_footer_readmore_page {
+  background-color: white;
+  padding: 10px;
+  margin-top: 10px;
+  margin-bottom: 10px;
+  padding-top: 11px;
```
niedzielski: Consider folding this into `padding` a couple lines up as `padding: 11px 10px;`.

https://developer.mozilla.org/en-US/docs/Web/CSS/padding#Syntax

src/transform/FooterReadMore.css
niedzielski: :+1: Thanks for checking!

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,91 @@
...
+    .forEach(element => {
+      element.style.backgroundPosition = `${rightOrLeft} ${margin}px center`
```
niedzielski: Do these need to change dynamically or can these inline styles be replaced with a platform style sheet (no JavaScript needed)?

```css
.pagelib-platform-android .pagelib_footer_menu_item {
  padding-left: 4px;
  padding-right: 4px;
}

.pagelib-platform-ios .pagelib_footer_menu_item {
  padding-left: 10px;
  padding-right: 10px;
}

.pagelib-platform-android .content-rtl .pagelib_footer_menu_item {
  background-position: 4px center;
}

.pagelib-platform-ios .content-rtl .pagelib_footer_menu_item {
  background-position: 10px left;
}
```

The only JavaScript you would need is to examine the user agent and could be used by all future transforms. Something like:

```js file=PlatformTransform.js
const PLATFORM = { ANDROID: 'pagelib-platform-android', IOS:  'pagelib-platform-ios' }

const isAndroid = window => /android/i.test(window.navigator.userAgent)
const isIOs = window => /ipad|iphone|ipod/i.test(window.navigator.userAgent)

const classify = window => {
  const html = window.document.querySelector('html')
  if (isAndroid(window)) { html.classList.add(PLATFORM.ANDROID) }
  if (isIOs(window)) { html.classList.add(PLATFORM.IOS) }
}

export default { PLATFORM, classify }
```
I haven't tested this so it probably needs some tweaking to actually work. The regexs are from [mobfro](https://phabricator.wikimedia.org/diffusion/EMFR/browse/master/resources/mobile.startup/browser.js).

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,218 @@
...
+ * @return {!DocumentFragment}
+ */
+const documentFragmentForMenuItem = (menuItem, document) => {
+  const item = document.createElement('div')
+  item.className = 'pagelib_footer_menu_item'
+
+  const containerAnchor = document.createElement('a')
+  containerAnchor.addEventListener('click', () => {
+    menuItem.clickHandler(menuItem.payload)
+  }, false)
```
niedzielski: `useCapture` defaults to false.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Parameters

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,218 @@
...
+    case MenuItemType.coordinate:
+      return 'pagelib_footer_menu_icon_coordinate'
+    default:
+      return ''
+    }
+  }
+
+  /**
+   * Returns reference to function for extracting payload when this menu item is tapped.
+   * @return {FooterMenuItemPayloadExtractor}
```
niedzielski: This should be `{?FooterMenuItemPayloadExtractor}` since the returned value may be undefined.

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,218 @@
...
+      return 'pagelib_footer_menu_icon_disambiguation'
+    case MenuItemType.coordinate:
+      return 'pagelib_footer_menu_icon_coordinate'
+    default:
+      return ''
+    }
+  }
+
+  /**
+   * Returns reference to function for extracting payload when this menu item is tapped.
```
niedzielski: Would it make more sense to just go ahead and call the method here on the client's behalf? Then they don't have to worry about `undefined` checks before invoking the method:

```js
    ...
    case MenuItemType.lastEdited:
      return undefined
    case MenuItemType.pageIssues:
      return pageIssuesStringsArray(document)
    case MenuItemType.disambiguation:
      return disambiguationTitlesArray(document)
    ...
```

src/transform/FooterContainer.js
```diff
@@ -0,0 +1,91 @@
...
+  </div>`
+  return containerFragment
+}
+
+/**
+ * Indicates whether container is has already been added.
+ * @param {!Document} document
+ * @return {boolean}
+ */
+const isContainerAttached = document => document.querySelector('#pagelib_footer_container') !== null
```
niedzielski: Can we use `Boolean(document.querySelector('#pagelib_footer_container'))` to avoid `null`?

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,218 @@
...
+ * @param {!string} containerID
+ * @param {FooterMenuItemClickCallback} clickHandler
+ * @param {!Document} document
+ * @return {void}
+ */
+const maybeAddItem = (title, subtitle, itemType, containerID, clickHandler, document) => {
+  const item = new MenuItem(title, subtitle, itemType, clickHandler)
+
+  // Items are not added if they have a payload extractor which fails to extract anything.
+  if (item.payloadExtractor() !== undefined) {
```
niedzielski: I think we can use the fact that a defined value will be truthy to simplify this to `if (item.payloadExtractor())`.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,337 @@
...
+  return stringToClean
+}
+
+/**
+ * Read more page model.
+ */
+class ReadMorePage {
+  /**
+   * ReadMorePage constructor.
+   * @param {!string} title
```
niedzielski: This is worth documenting that it is URL encoded.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,337 @@
...
+  const innerDivContainer = document.createElement('div')
+  innerDivContainer.classList.add('pagelib_footer_readmore_page_container')
+  outerAnchorContainer.appendChild(innerDivContainer)
+  outerAnchorContainer.href = `/wiki/${encodeURI(readMorePage.title)}`
+
+  if (readMorePage.title) {
+    const title = document.createElement('div')
+    title.id = index
+    title.className = 'pagelib_footer_readmore_page_title'
+    const displayTitle = readMorePage.title.replace(/_/g, ' ')
```
niedzielski: Do we need to call `decodeURI()` first? If so, this should probably live in a separate testable method.

src/transform/FooterReadMore.js
niedzielski: Er, not URL encoded :]

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,337 @@
...
+
+/**
+ * URL for retrieving 'Read more' pages for a given title.
+ * Leave 'baseURL' null if you don't need to deal with proxying.
+ * @param {!string} title
+ * @param {?string} baseURL
+ * @return {!sring}
+ */
+const readMoreQueryURL = (title, baseURL) => {
+  const readMoreItemFetchCount = 3
```
niedzielski: I like that you've made a nice symbol for this value but also consider exposing it as a parameter.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,337 @@
...
+  const readMoreQueryParameterString = stringFromQueryParameters(
+    queryParameters(title, readMoreItemFetchCount)
+  )
+  let baseURLToUse = baseURL
+  if (baseURLToUse === null) {
+    baseURLToUse = ''
+  }
+  return `${baseURLToUse}/w/api.php?${readMoreQueryParameterString}`
```
niedzielski: I think this can be simplified to `${baseURL || ''}/w/api.php?${readMoreQueryParameterString}`

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,337 @@
...
+ * @param {SaveButtonClickHandler} saveButtonClickHandler
+ * @param {TitlesShownHandler} titlesShownHandler
+ * @param {!Document} document
+ * @return {void}
+ */
+const fetchReadMore = (title, containerID, baseURL, showReadMorePagesHandler,
+  saveButtonClickHandler, titlesShownHandler, document) => {
+  const xhr = new XMLHttpRequest() // eslint-disable-line no-undef
+  xhr.open('GET', readMoreQueryURL(title, baseURL), true)
+  xhr.onload = function() {
```
niedzielski: Why not use addEventListener()?

niedzielski: Can we use lambdas?

montehurd: Fixed!

src/transform/FooterReadMore.js
montehurd: Oh good point. Fixed!

src/transform/FooterContainer.js
montehurd: Opps! Fixed!

src/transform/FooterReadMore.css
montehurd: CSS is so weird :) Fixed.

src/transform/FooterMenu.js
montehurd: Fixed!

src/transform/FooterMenu.js
montehurd: Fixed!

src/transform/FooterContainer.js
montehurd: Much better! Fixed!

src/transform/FooterMenu.js
montehurd: Fixed! Have I mentioned how much I hate JS's truthy voodoo? ;)

src/transform/FooterReadMore.js
montehurd: Fixed!

src/transform/FooterMenu.js
montehurd: The `undefined` check is needed I think. If an item doesn't define a payload extractor then we want to show that item - i.e. not return early before `addItem` is called.

montehurd: However, items which **do** define a payload extractor are **not** shown if no payload items are extracted.

src/transform/FooterContainer.js
montehurd: > Do these need to change dynamically or can these inline styles be replaced with a platform style sheet (no JavaScript needed)?

I think dynamically. 

The footer needs to always use the same side-margin as the article content. But on tablet we adjust the article content side margin (as you can see in the last screenshot below) when the table of contents is hidden.

screen shot 2017-07-17 at 3 19 33 pm

screen shot 2017-07-17 at 3 19 57 pm

screen shot 2017-07-17 at 3 20 08 pm

screen shot 2017-07-17 at 3 21 09 pm

screen shot 2017-07-17 at 3 21 19 pm

screen shot 2017-07-17 at 3 21 30 pm

montehurd: ^ especially note how in last screenshot above the `About this article` menu items extend all the way to the screen edge, yet the menu icons are following the article content edge - i.e. the icons left side sits just to right of the orange line . Design really wanted this.

I couldn't get all these requirements to play nice with pure CSS. (may be possible though)

montehurd: Reminder: If you search for `contentWidthPercentage` in the iOS codebase you can see what effects the margins.

montehurd:
@niedzielski 

here's the native [getCurrentVisibleFooterIndex](https://github.com/wikimedia/wikipedia-ios/pull/1615/files#diff-38568396230724f82d0bf5ec47cec4d8R834) method I mentioned. used so the native TOC can know what section should be selected

(the footer version of [getCurrentVisibleSection](https://github.com/wikimedia/wikipedia-ios/blob/develop/Wikipedia/Code/WebViewController.m#L814))

montehurd:
Oh and here's the [file](https://github.com/wikimedia/wikipedia-ios/blob/b9602116e333bc46ba85470d2bad97135296b5d7/Wikipedia/Code/WKWebView%2BWMFWebViewControllerJavascript.swift) that contains a lot of the native code which adds footer bits.

src/transform/FooterContainer.css
```diff
@@ -0,0 +1,24 @@
...
+    background-color: #eaecf0;
+}
+
+.pagelib_footer_container_heading {
+    padding-top: 35px;
+    font-size: 0.95em;
+    line-height: 1.6;
+}
+
+#pagelib_footer_container_readmore_pages {
```
niedzielski: Is it intentional to omit an underscore between `read` and `more`?

src/transform/FooterLegal.css
```diff
@@ -0,0 +1,38 @@
+
```
niedzielski: Some funky spacing here :] Extra newline at top and at the start and finish of the rule body for `.pagelib_footer_legal_contents`.

src/transform/FooterLegal.css
```diff
@@ -0,0 +1,38 @@
...
+}
+
+html[dir="ltr"] .pagelib_footer_legal_contents {
```
niedzielski: I don't mind if you want to use quotes but can we stick with single quotes? We currently use single quotes in all the JavaScript.

src/transform/FooterLegal.css
```diff
@@ -0,0 +1,38 @@
...
+
+.pagelib_footer_legal_license {
+  color: #54595d;
+  font-size: 0.75em;
+  line-height: 1.5;
+  display: block;
+}
+
+.pagelib_footer_legal_license_link {
+  color: #2C5BC5;
```
niedzielski: Would you mind lowercasing this color to match the other rules?

src/transform/FooterMenu.css
```diff
@@ -0,0 +1,63 @@
+
+.pagelib_footer_menu_item {
+    padding: 10px;
```
niedzielski: Would you mind if we stuck with two-space indentation for CSS too?

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,219 @@
...
+ * @typedef {function} FooterMenuItemClickCallback
+ * @param {!Array.<string>} payload Important - should return empty array if no payload strings.
+ * @return {void}
+ */
+
+/**
+ * @typedef {number} MenuItemType
+ */
+
+// eslint-disable-next-line
```
niedzielski: Please limit this to the rules you want to disable. e.g., `eslint-disable-next-line valid-jsdoc`.

src/transform/FooterMenu.js
```diff
@@ -0,0 +1,219 @@
...
+const pageIssuesStringsArray = document => {
+  const tables = Polyfill.querySelectorAll(document,
+    'div#content_block_0 table.ambox:not(.ambox-multiple_issues):not(.ambox-notice)'
+  )
+  // Get the tables into a fragment so we can remove some elements without triggering a layout
+  const fragment = document.createDocumentFragment()
+  for (let i = 0; i < tables.length; i++) {
+    fragment.appendChild(tables[i].cloneNode(true))
+  }
+  // Remove some element so their text doesn't appear when we use "innerText"
```
niedzielski: Because you're using a DocumentFragment, you have to manually walk clone and walk this list to remove hidden elements? Would using `window.document` allow you to avoid the clone and selective deletion? According to the MDN documentation, `Node.innerText` (as opposed to `Node.textContent`): "`innerText` is aware of style and will not return the text of hidden elements, whereas `textContent` will."

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

src/transform/FooterMenu.js
niedzielski: How about:

```js
const maybeAddItem = (title, subtitle, itemType, containerID, clickHandler, document) => {
  const item = new MenuItem(title, subtitle, itemType, clickHandler)
  item.payload = item.payloadExtractor(document)

  if (item.payload.length) {
    addItem(item, containerID, document)
  }
}
...
  /** @return {!Array.<string>} */
  payloadExtractor {
    switch (this.itemType) {
      case MenuItemType.pageIssues: return pageIssuesStringsArray
      case MenuItemType.disambiguation: return disambiguationTitlesArray
      default: return []
    }
  }
```

src/transform/FooterReadMore.css
```diff
@@ -0,0 +1,90 @@
+
+.pagelib_footer_readmore_page {
+  background-color: white;
+  padding: 11px 10px;
+  margin-top: 10px;
+  margin-bottom: 10px;
+  border-radius: 2px;
+  box-shadow: 0px 2px 4px rgba(200, 204, 209, 0.5);
+  display:block;
```
niedzielski: Missing a space here between `display` and `block` :]

src/transform/FooterReadMore.css
```diff
@@ -0,0 +1,90 @@
...
+    box-shadow: 0 0 6px rgba(200, 204, 209, 0.5);
+  }
+}
+
+.pagelib_footer_readmore_page_container {
+
+}
+
+.pagelib_footer_readmore_page_title {
+  color: #222222;
```
niedzielski: Would you mind if we used `#rgb[a]` notation here? `#222`

src/transform/FooterReadMore.css
```diff
@@ -0,0 +1,90 @@
...
...vau4IPcbamm+3ldXh8wcdTwOkKsPGgg8t2p1q3fnkK65alqOUL13op/LoOl3f0JEZFoNPpx1rq4GpDv1pX+OZUClaksmllKhtXPKXW/+lwUnDQgdMVeDwEWjt95JbYqW12KR3qJHauns4HOVPUNhuWJ1P01jS1fLWhH2upgy6nX6xAc7uX3BI7TW0eAAx6ib3r0lm1OHFI2/eWTmbX6unodMpNXNvsIm+/nfbuATEC9Tc85JbYae30AWAy6tiXn8lrC+KHveaNxYnsWZuO/rbEP60e8vbbudnhG1uB600u8kvt/NujZC/KpOPLzVm8PDv2gdcuWxjPF9YMDHpFQp3FW96xEaip7+fDAw66+5Tfb0y0nvJt2SyYOSHsPpbMiaO0IAuzURm+rctHXomduhZ3ZAUuXXeyuayWPreygsTFGKgoymZ25viHHviFWRMp25JFtFkPQEfvAPmlDv5qdEVG4NyVHrYeqsPtVYJPjDVSWWxhZtq4hw4+xDzLBA5vm0FMtCLR0+9n40EHNfX9YfcRlsCZ6i6Kj9TjG1AeQCkJZqp2WMhIiRpx8CGeyRhPRVE2cTHKrqbPHWBzWS0XrzvDuv6BAqfOd7CrqgF/QNlSpk2JomqHhalJ5kcOPsTMtHFUFltIjDUC4PYG2HaojnNXeh5N4MTZdvYcbSQQVILPTI3m22ILUyYZRy34EBkpSmKS400A+AaCFB+p50x118gEvj/dhu14E/LtzXwoSwkTI7IDB2Bqkvmu2fUHZHZVNXDqfMfDCVT9cpOyH5vV8rOZMVQUZRM7Xh+x4EMkx5uU+ys1GoBAUGbP0UZOnG0PT6D85A0qfr6hludZJlC+9c5KMRYkxhqp/ChbXeFkWcZ2vIkffm8b0lafOte6O1To9wQ5drpV/XLRrFgObMoiyjT2R+cok46c+fFUO/q41aVsNS5c60WSJKodd1Yoabi31195Lg7boEe+KNzeINvL6+46ZwzmnqnNmR/PvvxM4cEDRJt1lG2ZwaJZ995nDRF4/fkEPt8QmQP4SDEbJQ4UZPLqnEn3F1j1UhK716ajE5/4IRj0EjZrBssWJtxb4J0lk9n5fproOO+LXgd716Wz8sU7ByYdwJqcZIrfnjbijscSSYJP1kzn3aW3j6xHfmqRn1TKT7bIkixH8o+/yPMfGVAsFZFJcHoAAAAASUVORK5CYII=);
```
niedzielski: The indent is inconsistent here.

src/transform/FooterReadMore.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.pagelib_footer_readmore_page_save {
+  color: #2C5BC5;
```
niedzielski: One more lowercase case :]

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,336 @@
...
+ * @typedef {function} ShownReadMorePagesHandler
+ * @param {!Array.<object>} pages
+ * @param {!string} containerID
+ * @param {SaveButtonClickHandler} saveButtonClickHandler
+ * @param {TitlesShownHandler} titlesShownHandler
+ * @param {!Document} document
+ * @return {void}
+ */
+
+const _saveButtonIDPrefix = 'readmore:save:'
```
niedzielski: This looks like a static immutable constant, would you mind if we stuck with `UPPER_CASE`-like syntax for these. So, SAVE_BUTTON_ID_PREFIX? I guess our functions are also constants but this is constant data.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,336 @@
...
+    title.innerHTML = displayTitle
+    outerAnchorContainer.title = displayTitle
+    innerDivContainer.appendChild(title)
+  }
+
+  let description
+  if (readMorePage.terms) {
+    description = readMorePage.terms.description[0]
+  }
+  if ((description === undefined || description.length < 10) && readMorePage.extract) {
```
niedzielski: This could be concisely written as `(description || '').length < 10` or `(!description || description.length < 10)`.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,336 @@
...
+          saveButtonClickHandler,
+          titlesShownHandler,
+          document
+        )
+      } else {
+        fetchErrorHandler(xhr.statusText)
+      }
+    }
+  }
+  xhr.onerror = e => {
```
niedzielski: The curly braces do not appear to add any value here and the parameter is unused.

src/transform/FooterReadMore.js
```diff
@@ -0,0 +1,336 @@
...
+ * Updates save button bookmark icon for saved state.
+ * @param {!HTMLDivElement} button
+ * @param {!boolean} isSaved
+ * @return {void}
+ */
+const updateSaveButtonBookmarkIcon = (button, isSaved) => {
+  const unfilledClass = 'pagelib_footer_readmore_bookmark_unfilled'
+  const filledClass = 'pagelib_footer_readmore_bookmark_filled'
+  button.classList.remove(unfilledClass)
+  button.classList.remove(filledClass)
```
niedzielski: Since add won't doubly add, consider reducing this to: `button.classList.remove(isSaved ? unfilledClass : filledClass)`.

https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

src/transform/FooterMenu.js
montehurd: iirc I tried that. I'll add some tests with the tricky cases I found then we can safely try alternatives.

src/transform/FooterMenu.js
montehurd: Hmm... I'll take a closer look tomorrow, but it doesn't look like that accounts for the cases I outlined...

src/transform/FooterLegal.css
montehurd: Thoughts on holding off on CSS linting until we enable CSS linting ;)

src/transform/FooterMenu.js
montehurd: Ah will do.

src/transform/FooterReadMore.js
montehurd: Not big on upper-casing WHYAREYOUSHOUTING but I like consistency so will do :)

src/transform/FooterReadMore.js
montehurd: Ooh nice! I'll try that.

src/transform/FooterReadMore.js
montehurd: Oh sweet good catch!!

src/transform/FooterContainer.css
montehurd: Yes, but I'm open to changing this - perhaps as a follow-on?

src/transform/FooterReadMore.js
montehurd: Turns out classList `remove` can be passed multiple class names. I used that rather than adding another ternary.

src/transform/FooterReadMore.js
montehurd: Fixed!

src/transform/FooterReadMore.js
montehurd: Fixed!

src/transform/FooterReadMore.js
montehurd: Fixed!

src/transform/FooterMenu.js
montehurd: Fixed!

src/transform/FooterLegal.css
montehurd: Fixed!

src/transform/FooterMenu.js
montehurd: I put the extractor in a const make it a little more obvious that in the case there's no extractor defined that we don't want to bail early.

src/transform/FooterMenu.js
montehurd: If it's ok I'd prefer to consider internals refactors in follow-ups. Preferably after I've had a chance to add a few more tests.

src/transform/FooterReadMore.js
montehurd: Added!

src/transform/FooterReadMore.css
montehurd: Fixed!

src/transform/FooterReadMore.css
montehurd: Fixed!

src/transform/FooterReadMore.css
montehurd: Fixed!

src/transform/FooterReadMore.css
montehurd: Fixed!

src/transform/FooterMenu.css
montehurd: Fixed!

src/transform/FooterLegal.css
montehurd: Fixed!

src/transform/FooterLegal.css
montehurd: I fixed the CSS linty bits.

niedzielski:
:+1: published as v4.3.0. Please update your integration patch to reference this version (and don't forget to unlink before building the bundle).

montehurd:
> Please update your integration patch to reference this version (and don't forget to unlink before building the bundle)

hehe that's got me a few times :)

PR #54, opened 2017-07-13
New: add theming and dark mode

niedzielski:
Bug: T164311

### DIFF
--- a/src/transform/ElementUtilities.js
+++ b/src/transform/ElementUtilities.js
@@ -1,5 +1,6 @@
 import Polyfill from './Polyfill'
 
+// todo: drop ancestor consideration and move to Polyfill.closest().
 /**
  * Returns closest ancestor of element which matches selector.
  * Similar to 'closest' methods as seen here:
@@ -20,9 +21,21 @@ const findClosestAncestor = (el, selector) => {
 }
 
 /**
+ * @param {?Element} element
+ * @param {!string} property
+ * @return {?Element} The inclusive first element with an inline style or undefined.
+ */
+const closestInlineStyle = (element, property) => {
+  for (let el = element; el; el = el.parentElement) {
+    if (el.style[property]) { return el }
+  }
+  return undefined
+}
+
+/**
  * Determines if element has a table ancestor.
  * @param  {!Element}  el   Element
- * @return {boolean}        Whether table ancestor of 'el' is found
+ * @return {!boolean}       Whether table ancestor of 'el' is found
  */
 const isNestedInTable = el => Boolean(findClosestAncestor(el, 'table'))
 
@@ -86,6 +99,7 @@ const copyDataAttributesToAttributes = (source, destination, attributes) => {
 export default {
   findClosestAncestor,
   isNestedInTable,
+  closestInlineStyle,
   isVisible,
   moveAttributesToDataAttributes,
   moveDataAttributesToAttributes,
--- a/src/transform/LazyLoadTransform.js
+++ b/src/transform/LazyLoadTransform.js
@@ -1,5 +1,6 @@
 import './LazyLoadTransform.css'
 import ElementUtilities from './ElementUtilities'
+import Polyfill from './Polyfill'
 
 // CSS classes used to identify and present converted images. An image is only a member of one class
 // at a time depending on the current transform state. These class names should match the classes in
@@ -190,8 +191,7 @@ const loadImage = (document, image) => {
  * @return {!HTMLImageElement[]} Convertible images descendent from but not including element.
  */
 const queryLazyLoadableImages = element =>
-  Array.prototype.slice.call(element.querySelectorAll('img'))
-    .filter(image => isLazyLoadable(image))
+  Polyfill.querySelectorAll(element, 'img').filter(image => isLazyLoadable(image))
 
 /**
  * Convert images with placeholders. The transformation is inverted by calling loadImage().
--- a/src/transform/LazyLoadTransform.js
+++ b/src/transform/LazyLoadTransform.js
@@ -202,4 +202,8 @@ const queryLazyLoadableImages = element =>
 const convertImagesToPlaceholders = (document, images) =>
   images.forEach(image => convertImageToPlaceholder(document, image))
 
-export default { loadImage, queryLazyLoadableImages, convertImagesToPlaceholders }
+export default {
+  loadImage,
+  queryLazyLoadableImages,
+  convertImagesToPlaceholders
+}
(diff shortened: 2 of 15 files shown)


src/transform/ThemeTransform.js
```diff
@@ -0,0 +1,51 @@
+import './ThemeTransform.css'
+import ElementUtilities from './ElementUtilities'
+import Polyfill from './Polyfill'
+
+// Elements marked with either of these classes indicate certain ancestry constraints that are
+// difficult to describe as CSS selectors.
+const CONSTRAINT = {
+  NO_BACKGROUND_IMAGE: 'pagelib-theme-image-no-background',
```
niedzielski: Rename IMAGE_WITHOUT_BACKGROUND.

niedzielski: Done

src/transform/ElementUtilities.js
```diff
@@ -20,9 +21,21 @@ const findClosestAncestor = (el, selector) => {
 }
 
 /**
+ * @param {?Element} element
+ * @param {!string} property
+ * @return {?Element} The inclusive first element with an inline style or undefined.
+ */
+const closestInlineStyle = (element, property) => {
+  for (let el = element; el; el = el.parentElement) {
```
montehurd: Do you need the middle `el;`? Wasn't sure if JS just lets you leave it blank...

niedzielski: Yes, the truthiness of `el` determines when the loop ends.

src/transform/LazyLoadTransform.js
```diff
@@ -190,8 +191,7 @@ const loadImage = (document, image) => {
  * @return {!HTMLImageElement[]} Convertible images descendent from but not including element.
  */
 const queryLazyLoadableImages = element =>
-  Array.prototype.slice.call(element.querySelectorAll('img'))
-    .filter(image => isLazyLoadable(image))
+  Polyfill.querySelectorAll(element, 'img').filter(image => isLazyLoadable(image))
```
montehurd: Thoughts on maybe extending prototype on HTMLElement with something like `wmf_querySelectorAll` so don't have to pass `element`? Was wondering about this earlier...

src/transform/Polyfill.js
```diff
@@ -17,6 +17,14 @@ const matchesSelector = (el, selector) => {
   return false
 }
 
+/**
+ * @param {!Element} element
+ * @param {!string} selector
+ * @return {!Element[]}
+ */
+const querySelectorAll = (element, selector) =>
```
montehurd: Thoughts on maybe extending prototype on HTMLElement with something like `wmf_querySelectorAll` so don't have to pass `element`? Was wondering about this earlier...

niedzielski: Me too. I'm not sure what's best practice for a library such as pagelib. I was thinking if we did it, we'd want to call it the same name, `querySelectorAll()`, which would have these pros and cons as I understand it:

Pros:
- Fewer missed polyfill mistakes.
- Less typing.
- Platform differences are better abstracted.

Cons:
- Potential incompatibility with other client libraries.

I say we punt on this for now until we feel strongly one way or another.

src/transform/ThemeTransform.js
```diff
@@ -0,0 +1,51 @@
...
+    if (!ElementUtilities.closestInlineStyle(image, 'background')) {
+      image.classList.add(CONSTRAINT.IMAGE_NO_BACKGROUND)
+    }
+    if (!ElementUtilities.isNestedInTable(image)) {
+      image.classList.add(CONSTRAINT.IMAGE_NONTABULAR)
+    }
+  })
+}
+
+export default { CONSTRAINT, THEME, setTheme, classifyElements }
```
montehurd: I've been using separate lines for these `export` statements. Thoughts on maybe doing same?

montehurd: Oh does `classifyElements` get called externally?

test/transform/CompatibilityTransform.test.js
```diff
@@ -0,0 +1,13 @@
+import assert from 'assert'
+import domino from 'domino'
+import pagelib from '../../build/wikimedia-page-library-transform'
+
+describe('CompatibilityTransform', () => {
+  it('.enableSupport()', () => {
+    const document = domino.createDocument()
+    pagelib.CompatibilityTransform.enableSupport(document)
+    for (const clazz of Object.values(pagelib.CompatibilityTransform.COMPATIBILITY)) {
```
montehurd: Did you mention these sorts of deliberate misspellings (`clazz`) are a pattern in JS?

They make my cheek twitch every time I see them.

src/transform/ElementUtilities.js
montehurd: Ah gotcha.

test/transform/CompatibilityTransform.test.js
niedzielski: Class is a keyword in ES6 just like C++. When I replace `clazz` with `class`, I get a parsing error in ESLint. Similarly, `thiz` or `that` is also frequently used to refer to a reference of `this`.

montehurd: I as just wondering if misspellings are the way to go vs something like `thisClass` or some other correctly spelled word(s).

montehurd: I may be a little OCD on this one ;) Having said that, I don't feel strongly about it.

niedzielski: I find `thisClass` to be a little distracting because what I'm really wanting is a simple singular of a plural, classes. I'm kind of anti-`myVariable` / `thisVariable` :] I didn't use `className` to avoid confusion with `Element.className`. I wouldn't mind `value` or `classValue` if you prefer it.

src/transform/ThemeTransform.js
niedzielski: `classifyElements` is called per blob of HTML (it's a transform) which is normally either the entire page's content or once per section. I'm open to rethinking how this transform is done but this seemed like the best choice for both platforms.

In general, I think I prefer to bunch object properties like these exports but I don't mind always keeping one export per line.

montehurd: Would you mind adding a line to its JSDoc explaining that this only needs to be called once, not each time you switch themes?

test/transform/CompatibilityTransform.test.js
montehurd: Value works for me.

src/transform/ThemeTransform.js
niedzielski: :+1: nice catch! I've updated this here and will push a fix to the Android app... :]

test/transform/CompatibilityTransform.test.js
niedzielski: :+1: done.

PR #55, opened 2017-07-14
Chore: require valid JSDocs

niedzielski:


montehurd:
Looks like our CSS whitespace isn't getting linted... is this something we can turn on too?

niedzielski:
Let's handled CSS linting in T163831.

PR #61, opened 2017-07-21
Fix: dark mode collapsed table secondary text and edit link colors

niedzielski:
Bug: T165415


niedzielski:
@montehurd, these are some fixes Rita identified in https://phabricator.wikimedia.org/T165415. This matches the current Android implementation and covers a fair bit of the future edit link consolidation task including zapping a sizable amount of code from the MobileApp extension. I know you have some concerns around keeping the interactivity of the buttons on Android but unless you feel strongly about them I'd like to punt for now so I can focus on getting the footer work completed.

src/transform/EditTransform.js
```diff
@@ -0,0 +1,34 @@
...
+const DATA_ATTRIBUTE = { SECTION_INDEX: 'data-id', ACTION: 'data-action' }
+const ACTION_EDIT_SECTION = 'edit_section'
+
+/**
+ * @param {!Document} document
+ * @param {!number} index The zero-based index of the section.
+ * @return {!HTMLSpanElement}
+ */
+const newEditSectionLink = (document, index) => {
+  const container = document.createElement('span')
```
montehurd: Thoughts on decomposing the span bits out of the link function? Specifically, adding a new `newEditSectionSpan` function and making it public, then it would call `newEditSectionLink` to set the span contents. In this scenario I'd make `newEditSectionLink` private (moving its export to a `test` object) and `newEditSectionSpan` public.

montehurd:
I noticed if you slowly resize the demo there are widths at which the RTL icon seems to jump out of its div a bit:

screen shot 2017-07-21 at 4 36 30 pm

May just be a demo artifact but thought it was worth asking about.

src/transform/EditTransform.js
niedzielski: I don't mind breaking the true link part out like you suggested but I think it's clearer to the client if we keep calling this a "link" or a "button." This makes a new component and that component is logically a link or a button. Thoughts?

montehurd: Oh that's why I tweaked the name of the span function above to newEditSection**Icon**Span...

montehurd: Something about calling it a link but returning a span seemed odd... I liked `icon` because it conveyed the button-ness... though I also like newEditSection**Button**Span too...

montehurd: If I had to choose between decomposing the 2 functions and the naming change I'd choose the decomposing. No strong feelings about the specific naming, just things to consider - whatever you choose is fine.

niedzielski: Thanks for the feedback! I've broken it into two methods, link and button. I'm still not quite happy with button since an image can be a link, for example, but I believe it captures the spirit adequately.

niedzielski:
@montehurd, thanks for the catch! I had to research and rework this a little bit to get it just right. Let me know if you're still seeing wrapping issues.

montehurd:
@niedzielski 

> Let me know if you're still seeing wrapping issues.

I don't see the issue anymore! Nice fix!

niedzielski:
Published in v4.5.0.

PR #66, opened 2017-07-24
Chore: add demo prototype framework

niedzielski:


montehurd:
I must be missing a setup step - when I run `npm run -s dev` and click through to the demo control panel page I'm getting JS errors...

demo/ControlPanel.html
```diff
@@ -0,0 +1,203 @@
...
+    initCallback(clone, '.theme', this.onThemeChange)
+    initCallback(clone, '#dim_images', this.onDimImagesChange)
+    initCallback(clone, '#compatibility_filter', this.onCompatibilityFilterChange)
+    initCallback(clone, '.page_protection', this.onPageProtectionChange)
+
+    const root = this.createShadowRoot()
+    root.appendChild(clone)
+  }
+
+  document.registerElement('control-panel', { prototype })</script>
```
montehurd: Looks like `registerElement` is what's hanging... looks like [it's deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/registerElement)?

montehurd:
Got the final thumbs-up for using the Android pencil on iOS - gonna hop back on integrating the EditTransform for a bit...

demo/ControlPanel.html
niedzielski: Weird. It definitely works on my machine but I definitely didn't know this was deprecated :| I've rewritten it using the new API. It's even rougher now... Please give it a shot! :] Oh and double check you've run `npm install` after checking this branch out!

montehurd: Cool I'll try it!

PR #69, opened 2017-07-25
Chore/browsersync configuration

niedzielski:


demo/build
```diff
@@ -0,0 +1 @@
+../build
```
niedzielski: This is a symlink.

PR #71, opened 2017-07-25
Fix: add missing href on edit anchor

montehurd:


niedzielski:
Published as NPM v4.5.3.

PR #72, opened 2017-07-25
Chore: add FooterTransformer and demo

niedzielski:


niedzielski:
@montehurd, I know this looks pretty gross. The reason I'm pushing this state is that I kind of needed a framework to build and test on that had most of what a client wanting to show a footer would need (it's still missing menu stuff among other things). I will be submitting more refactor PRs on top of this.

PR #76, opened 2017-07-27
Fix: add Web Components polyfill to demos

niedzielski:
Add Web Components polyfill to each demo. I couldn't find
webcomponents-loader.js as an NPM distributable so I used the one hosted
for Polymer projects.

https://github.com/webcomponents/webcomponentsjs
https://github.com/webcomponents/webcomponentsjs/issues/397
https://www.npmjs.com/package/webcomponents.js
https://www.polymer-project.org/2.0/start/quick-tour


niedzielski:
@montehurd, please try in Safari.

montehurd:
Hmm it seems to be stuck in some sort of redirect loop...

montehurd:
- safari : redirect loop?
- firefox: blank
- chrome: seems fine

niedzielski:
Sheesh, this doesn't work at all? WTF... What version of Firefox are you running? This seems to work correctly on Firefox v54.0 for Linux.

montehurd:
Looks like I was on 49.0.1. I updated and will re-test when I get a chance.

montehurd:
The demo html files start working if you add a `link` tag for 'wikimedia-page-library-transform.css' to each file. Only problem is the polyfill custom control tag is still doing some weird redirect loop on safari and firefox...

montehurd:
> The demo html files start working...

on Safari and Firefox that is

niedzielski:
@montehurd the style links are definitely missing! I'm disappointed that at least Firefox isn't working :[

montehurd:
Could it be refactored a bit to use a CSS and JS file instead of `<link rel=import href=ControlPanel.html>`?

niedzielski:
You could refactor the JavaScript and CSS easily but I'm not sure how you would inject the ControlPanel HTML into the page without extra tooling.

montehurd:
Given the compatibility issues, do you think we should consider more vanilla JS ways of adding control panel elements to demo pages? I think the way you did it is super cool though :)

niedzielski:
I'm fine with exploring other options like iframe.

niedzielski:
@montehurd, please try once more.

montehurd:
Tested on Safari, Chrome and Firefox.

niedzielski:
Thank heavens! No it was just WIP cause I thought it was working before and then it wasn't and I couldn't really verify on my machine.

PR #77, opened 2017-07-28
Fix: Various theme issues. Also added theme transform demo.

montehurd:
https://phabricator.wikimedia.org/T171326

Article | Before | After
--- | --- | ---
Mycenaean Greek |  | 
Mycenaean Greek |  | 
Mycenaean Greek |  | screen shot 2017-08-10 at 1 12 34 pm
John McCain |  | 
Chris Cornell |  | screen shot 2017-08-10 at 1 14 34 pm
Chris Cornell |  | screen shot 2017-08-10 at 1 15 42 pm
Kill 'Em All |  | 
Audioslave |  | screen shot 2017-08-10 at 1 17 23 pm
Audioslave |  | screen shot 2017-08-10 at 1 19 04 pm
Manchester United F.C. |  | screen shot 2017-08-10 at 1 21 54 pm
Toast |  | screen shot 2017-08-10 at 1 22 46 pm
Mathematics |  | screen shot 2017-08-10 at 1 23 36 pm
Gantz | screen shot 2017-08-10 at 1 40 59 pm | screen shot 2017-08-10 at 1 36 50 pm
Game of Thrones (season 7) | screen shot 2017-08-10 at 1 53 14 pm | screen shot 2017-08-10 at 1 56 49 pm
Mycenaean Greek | screen shot 2017-07-28 at 3 58 40 pm | screen shot 2017-07-28 at 4 05 21 pm
Mycenaean Greek | screen shot 2017-07-28 at 3 59 39 pm | screen shot 2017-07-28 at 4 05 37 pm
John McCain | screen shot 2017-07-28 at 4 00 11 pm | screen shot 2017-07-28 at 4 05 58 pm
Chris Cornell | screen shot 2017-07-28 at 4 00 44 pm | screen shot 2017-08-10 at 1 25 59 pm
Chris Cornell | screen shot 2017-07-28 at 4 01 03 pm | screen shot 2017-08-10 at 1 27 14 pm
Kill 'Em All | screen shot 2017-07-28 at 4 01 25 pm | screen shot 2017-07-28 at 4 08 56 pm
Audioslave | screen shot 2017-07-28 at 4 01 50 pm | screen shot 2017-08-10 at 1 30 33 pm
Audioslave | screen shot 2017-07-28 at 4 02 13 pm | screen shot 2017-08-10 at 1 31 29 pm
Manchester United F.C. | screen shot 2017-07-28 at 4 02 44 pm | screen shot 2017-08-10 at 1 32 35 pm
Toast | screen shot 2017-07-28 at 4 03 07 pm | screen shot 2017-08-10 at 1 33 37 pm
Mathematics | screen shot 2017-07-28 at 4 03 30 pm | screen shot 2017-08-10 at 1 34 33 pm
Gantz | screen shot 2017-08-10 at 1 40 52 pm | screen shot 2017-08-10 at 1 36 41 pm
Game of Thrones (season 7) | screen shot 2017-08-10 at 1 53 31 pm | screen shot 2017-08-10 at 1 56 42 pm


montehurd:
Article | Before | After
--- | --- | ---
Mycenaean Greek |  | 
Mycenaean Greek |  | 
Mycenaean Greek |  | 
John McCain |  | 
Chris Cornell |  | 
Chris Cornell |  | 
Kill 'Em All |  | 
Audioslave |  | 
Audioslave |  | 
Manchester United F.C. |  | 
Toast |  | 
Mathematics |  |

montehurd:
I've spent a little time refactoring the theme demo file... will push an update soon-ish - though it could come as a follow-on as well :)

demo/DemoArticles/en.Audioslave.json
```diff
@@ -0,0 +1 @@
...e","anchor":"Musical_style_and_influences","id":9,"text":"<p>By combining '70s style hard rock riffing with alternative rock, Audioslave created a distinctive sound.<sup id=\"cite_ref-Plugged_In_Online_-_Audioslave_Comes_.27Out_of_Exile.27_103-0\" class=\"reference\"><a href=\"#cite_note-Plugged_In_Online_-_Audioslave_Comes_.27Out_of_Exile.27-103\">[103]</a></sup> This mix was driven by Cornell's wide vocal range,<sup id=\"cite_ref-104\" class=\"reference\"><a href=\"#cite_note-104\">[104]</a></sup> Morello's innovative guitar solos and the robust rhythm section of Wilk and Commerford.<sup id=\"cite_ref-105\" class=\"reference\"><a href=\"#cite_note-105\">[105]</a></sup> Morello, although stating he \"never felt musically limited\" in Rage Against the Machine, did say that he had \"a lot more scope to explore with Audioslave\" and a \"wider musical territory.\"<sup id=\"cite_ref-106\" class=\"reference\"><a href=\"#cite_note-106\">[106]</a></sup> This meant that the instrumentalists had the opportunity to write slow and melodic songs, something they had not done before.</p>\n<p>As opposed to de la Rocha's lyrics, Cornell's were mostly apolitical; Morello referred to them as \"haunt
```
niedzielski: Since this is nonproduction code, please format these JSON samples for easy reading. I use [a bookmarklet](https://gist.github.com/niedzielski/53c98af986955053aaabcc8a19a62f39) when viewing responses in browser but there are many other ways to go about it.

demo/ThemeTransform.html
```diff
@@ -0,0 +1,109 @@
...
+      :root {
+        --control-panel-theme-visibility: visible;
+      }
+      body{
+        padding: 20px 20px 20px 300px !important;
+      }
+      .content_block {
+        padding-bottom: 50px;
+      }
+      #demoMenu a {
```
niedzielski: Please use underscores for CSS identifiers per earlier discussion.

src/transform/ThemeTransform.css
```diff
@@ -35,17 +35,121 @@
   background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAADsMAAA7DAcdvqGQAAAAadEVYdFNvZnR3YXJlAFBhaW50Lk5FVCB2My41LjEwMPRyoQAAAGJJREFUKFN1jdENgCAMBYmJn47Bak7DZrhTpc/XIm34OAjXA4qIgHI/dSBbLGTcOKjBryFlinGmjDQGiOF0MQkxI3v5wq6L38qR7SnsAx8ul37igPjAd+o5Oz2MRA+xY4ZSXuaW6wYouOLpAAAAAElFTkSuQmCC);
 }
 
+.pagelib-theme-dark .content table.mbox-small{
+  color: #f8f9fa !important; 
+}
+
+.pagelib-theme-sepia .content table.mbox-small{
```
niedzielski: Missing space before bracket.

src/transform/ThemeTransform.css
```diff
@@ -35,17 +35,121 @@
   background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAADsMAAA7DAcdvqGQAAAAadEVYdFNvZnR3YXJlAFBhaW50Lk5FVCB2My41LjEwMPRyoQAAAGJJREFUKFN1jdENgCAMBYmJn47Bak7DZrhTpc/XIm34OAjXA4qIgHI/dSBbLGTcOKjBryFlinGmjDQGiOF0MQkxI3v5wq6L38qR7SnsAx8ul37igPjAd+o5Oz2MRA+xY4ZSXuaW6wYouOLpAAAAAElFTkSuQmCC);
 }
 
+.pagelib-theme-dark .content table.mbox-small{
```
niedzielski: This was all ported code but I still tried to add examples as I went. Since this is new stuff and you've already identified the articles, please add them as comments!

src/transform/ThemeTransform.css
```diff
@@ -35,17 +35,121 @@
   background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAADsMAAA7DAcdvqGQAAAAadEVYdFNvZnR3YXJlAFBhaW50Lk5FVCB2My41LjEwMPRyoQAAAGJJREFUKFN1jdENgCAMBYmJn47Bak7DZrhTpc/XIm34OAjXA4qIgHI/dSBbLGTcOKjBryFlinGmjDQGiOF0MQkxI3v5wq6L38qR7SnsAx8ul37igPjAd+o5Oz2MRA+xY4ZSXuaW6wYouOLpAAAAAElFTkSuQmCC);
```
niedzielski: It appears that every new style added in this file has important priority. Is this necessary? On one hand, this means we have a pretty good chance that theming will work. On the other, this may be just as likely to break elements that were intentionally styled differently and now would have to battle this new pagelib `important` with another important.

demo/ThemeTransform.html
```diff
@@ -0,0 +1,109 @@
+<!DOCTYPE html>
```
niedzielski: This demo is crazy useful!!

demo/ThemeTransform.html
```diff
@@ -0,0 +1,109 @@
...
+    'en.Kill_\'Em_All.json',
+    'en.Manchester_United_F.C..json',
+    'en.John_McCain.json',
+    'en.Playwright.json',
+    'en.Toast.json'
+  ]
+
+  const urls = fileNames.map(filename => `/DemoArticles/${filename}`)
+
+  const demoMenu = document.getElementById('demoMenu')
```
niedzielski: Consider using a drop down here to match all the demos and providing both Parsoid and MediaWiki response types since Android uses Parsoid (via the Content Service) whenever possible.

demo/ThemeTransform.html
```diff
@@ -0,0 +1,109 @@
...
+        width: 230px;
+        border-radius: 8px;
+      }
+      .demoArticleTitle {
+        color: red;
+        font-size: 2.0em;
+      }
+    </style>
+ 
+    <link href="https://en.wikipedia.org/w/load.php?debug=false&lang=en&only=styles&skin=vector&modules=skins.minerva.base.reset|skins.minerva.content.styles|ext.math.styles|" rel="stylesheet" type="text/css"></link>
```
niedzielski: This ends with a closing pipe which seems incorrect. Android uses the following styles: `?debug=true&lang=en&modules=skins.minerva.base.reset|skins.minerva.content.styles|ext.cite.style|ext.math.styles|ext.timeline.styles|mediawiki.page.gallery.styles|mobile.app.pagestyles.android|mediawiki.skinning.content.parsoid&only=styles&version=&*`. I like to keep `debug=true` on for demos unless performance is bad. `mobile.app.pagestyles.android` is probably incorrect in the long term and we should probably update our usage to include more comprehensive styling from Minerva Neue, MediaWiki:*, and the extensions some day (but not today).

https://www.mediawiki.org/wiki/Manual:Interface/Stylesheets

demo/ThemeTransform.html
```diff
@@ -0,0 +1,109 @@
...
+      <div id='demoMenu'></div>
+    </div>
+    <div class='content' id='content'></div>
+    
+    <script>
+        
+(function() {
+
+  // The json files below were saved using the mobileview parameters in the URL below.
+  // eslint-disable-next-line max-len
```
niedzielski: Please just disable the rule for URLs in the ESLint config. I've found the following works well: max-len: `[1, {tabWidth: 2, ignoreUrls: true, ignorePattern: '^/// .+'}]`.

demo/ThemeTransform.html
```diff
@@ -0,0 +1,109 @@
...
+    titleAnchor.innerHTML = url
+    demoMenu.appendChild(titleAnchor)
+  })
+
+  // Layout the article section data consecutively on a single page.
+  // Comment out any files above you aren't interested in testing, then when finished double-check
+  // all of them.
+  Promise.all(urls.map(url => fetch(url).then(resp => resp.json())))
+    .then(jsons => jsons.map(json => json.mobileview.sections))
+    .then(sections => [].concat(...sections))
```
niedzielski: Why is a copy needed?

demo/ThemeTransform.html
```diff
@@ -0,0 +1,109 @@
...
+                                <h2>${section.line}</h2>
+                                <br>${section.text}
+                              </div>`
+      )
+    )
+    .then(sections => sections.join(''))
+    .then(text => {
+      document.querySelector('#content').innerHTML = text
+
+      Array.from(document.querySelectorAll('.content_block[id="0"] h2:first-of-type'))
```
niedzielski: Is Array.from() needed in the newer browsers we're targeting? Won't there only be a single result anyway (can we use `querySelector()` instead)?

demo/ThemeTransform.html
```diff
@@ -0,0 +1,109 @@
...
+    const titleAnchor = document.createElement('a')
+    titleAnchor.href = `#${url}`
+    titleAnchor.innerHTML = url
+    demoMenu.appendChild(titleAnchor)
+  })
+
+  // Layout the article section data consecutively on a single page.
+  // Comment out any files above you aren't interested in testing, then when finished double-check
+  // all of them.
+  Promise.all(urls.map(url => fetch(url).then(resp => resp.json())))
```
niedzielski: Just curious what the reason is to load these all at once instead of as they're selected.

src/transform/ThemeTransform.css
```diff
@@ -35,17 +35,121 @@
...
+}
+
+.pagelib-theme-sepia .content table.mbox-small-left hr,
+.pagelib-theme-sepia .content table.mbox-small hr {
+  border-style: solid !important;
+  background: #cbc8c1 !important;
+  border-color: #cbc8c1 !important;
+}
+
+.pagelib-theme-dark .content div.noprint.portal.plainlist > ul {
```
niedzielski: Is it intentional to limit this rule to direct children of `div.noprint.portal.plainlist`?

src/transform/ThemeTransform.css
montehurd: Ah, I will double-check these.

demo/DemoArticles/articles.json
```diff
@@ -0,0 +1,14 @@
+[
```
niedzielski: This looks nice! Please add a field for the backend used too!

demo/DemoArticles/update.js
```diff
@@ -0,0 +1,33 @@
...
+  const url = urlFromArticleData(articleData)
+  request({
+    method: 'GET',
+    uri: url,
+    gzip: true
+  }, (error, response, body) => {
+    if (!error && response.statusCode === 200) {
+      // eslint-disable-next-line no-console
+      console.log(`\tJSON saved to '${fileName}'`)
+      fs.writeFile(`./data/${fileName}`, body, err => { if (err) { throw err } })
```
niedzielski: Please format the response using something like `JSON.stringify(JSON.parse(body), null, 2);`

demo/DemoArticles/articles.json
montehurd: Will do!

src/transform/ThemeTransform.css
montehurd: Will do!

demo/DemoArticles/en.Audioslave.json
montehurd: Fixed!

demo/ThemeTransform.html
montehurd: I did but hadn't pushed in time :)

montehurd:
@niedzielski Hmm... I added a Hebrew test file in the last commit and it's tripping that weird control character linting warning again... 

Good news though - I was able to greatly simplify the CSS while at the same time making it provide better theme coverage for non-EN langs.

montehurd:
This [commit](2b24f9c) was potentially interesting - It looks like I was able to greatly simplify the CSS while at the same time making it provide better theme coverage for non-EN langs - and especially on parts using inline styles. Went from having quite a few remaining un-themed elements to none, as far as I can tell, even after adding a couple additional demo articles. I was just wondering how far I could get by being agnostic about css class names and started playing with it, scrolling through my giant page of articles to see what worked.

Tomorrow I'd like to add at least one theme demo article for each the remaining top-ten wiki langs (I'll try to find really long articles) and see if I can identify any outstanding theme issues.

montehurd:
I'll also rebuild the 'after' screenshots today...

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,25 @@
+/*
+#!/usr/bin/env node
+*/
+/* eslint-disable require-jsdoc */
+
+class ArticleRef {
+  constructor(lang, title, revision) {
+    this.lang = lang
```
niedzielski: I like the way you've restructured this as a type. Consider indicating that these variables are private by using a leading underscore.

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+  displayName() {
+    return `"${this.lang} > ${this.title}"`
+  }
+  url() {
+    return `https://${this.lang}.wikipedia.org/w/api.php?action=mobileview&page=${encodeURI(this.title)}&revision=${this.revision}&format=json&noheadings=true&prop=sections%7Ctext%7Clastmodified%7Clastmodifiedby%7Clanguagecount%7Cid%7Cprotection%7Ceditable%7Cdisplaytitle%7Cthumb%7Cdescription%7Cimage%7Crevision%7Cnamespace&sectionprop=toclevel%7Cline%7Canchor%7Clevel%7Cnumber%7Cfromtitle%7Cindex&sections=all&thumbwidth=640`
+  }
+}
+
+module.exports = {
```
niedzielski: Not a storng preference but you can also just do `export default class { ... }`.

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,25 @@
+/*
+#!/usr/bin/env node
+*/
+/* eslint-disable require-jsdoc */
```
niedzielski: I recommend adding simple typing docs like `@return {!string}` and nothing more.

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,25 @@
+/*
+#!/usr/bin/env node
```
niedzielski: I believe this must appear on the first line or the shebang has no effect and will probably use the default interpreter, mostly likely `sh`.

https://en.wikipedia.org/wiki/Shebang_(Unix)

demo/DemoArticles/articles.json
```diff
@@ -0,0 +1,24 @@
+[ 
+  {"lang": "en", "title": "Barack Obama", "revision": 793008506},
```
niedzielski: Any reason not to use ArticleRef for each entry here?

src/transform/ThemeTransform.css
```diff
@@ -35,51 +33,71 @@
...
+  background: #27292d !important;
+}
+.pagelib-theme-sepia .content div.quotebox {
   background: #f0e6d6 !important;
 }
 
-.pagelib-theme-dark li.gallerybox div.thumb {
-  border-color: #43464a;
-  background: #27292d;
+.pagelib-theme-dark .content * {
```
niedzielski: These wildcard importants look a little dangerous. Have you noticed any regressions?

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,25 @@
...
+/* eslint-disable require-jsdoc */
+
+class ArticleRef {
+  constructor(lang, title, revision) {
+    this.lang = lang
+    this.title = title
+    this.revision = revision
+  }
+  fileName() {
+    return `${this.lang}.${this.title}.${this.revision}.json`
```
niedzielski: Should we include the revision in the filename? If we ever change the revision, we'll see a new file generated and the old one will be unchanged and not identified by Git.

src/transform/ThemeTransform.css
```diff
@@ -35,51 +33,71 @@
...
+.pagelib-theme-dark .content div,
+.pagelib-theme-dark .content span,
+.pagelib-theme-dark .content ul,
+.pagelib-theme-dark .content li {
+  background: transparent !important;
+  color: #f8f9fa !important;
+}
+.pagelib-theme-sepia .content div,
+.pagelib-theme-sepia .content span,
+.pagelib-theme-sepia .content ul,
```
niedzielski: Should we do `ol` too?

demo/DemoArticles/ArticleRef.js
niedzielski: Actually, it appears that this file is not meant to be executed so I think the shebang should be removed (and if the executable bit is set, that should be removed too).

demo/DemoArticles/update.js
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env node
+/* eslint-disable require-jsdoc */
+
+const fs = require('fs')
+const request = require('request')
+const INDEX_FILE = './articles.json'
+const ArticleRef = require('./ArticleRef.js').ArticleRef
+
+const articleRefs = JSON
+  .parse(fs.readFileSync(INDEX_FILE, 'utf8'))
```
niedzielski: Can't you just `require` a JSON file?

demo/ThemeTransform.html
```diff
@@ -0,0 +1,134 @@
...
+
+  <body>
+    <div id='theme_demo_tools'>
+      <control-panel></control-panel>
+      <div id='theme_demo_menu'></div>
+    </div>
+    <div class='content' id='content'></div>
+    
+    <script>
+/* eslint-disable require-jsdoc */
```
niedzielski: I really recommend using JSDoc typing when Promises are used.

montehurd: Added!

demo/DemoArticles/update.js
montehurd: Oh god good catch!

montehurd: Fixed!

src/transform/ThemeTransform.css
montehurd: Added!

demo/DemoArticles/ArticleRef.js
montehurd: Yes I believe we should. The reason is we may want to test 2 revisions of the same article.

src/transform/ThemeTransform.css
montehurd: Agreed these are _usually_ not recommended - it's a blunt tool.

The reason I switched to this 'baseline' approach:

There were many places, especially in non-EN wikis, where inline styles are used - often on elements having _no css classes_ specified. Without having CSS classes, it was almost impossible to identify and target such elements for theming.

This approach handles such elements and provide consistent theming.

The fallout from this approach:
- simplified other theme transform CSS - much of it no longer needed
- theme CSS needs to come before other transform CSS in the file delivered by the package (otherwise the other transforms wouldn't be able to override theme CSS if they needed to - note when doing so these transforms need to target elements using a >= level of specificity as used by the theme transform)

demo/DemoArticles/articles.json
montehurd: Hmm I'm open to it, but not sure what that would add...

demo/DemoArticles/ArticleRef.js
montehurd: Fixed!

demo/DemoArticles/ArticleRef.js
montehurd: For some reason I can't get it working in this file. Maybe it's not being interpreted as ES6?

demo/DemoArticles/ArticleRef.js
montehurd: They're public.

demo/DemoArticles/update.js
montehurd: This was already fixed.

src/transform/ThemeTransform.css
montehurd: No longer present after the 'baseline' refactor.

demo/ThemeTransform.html
montehurd: Proofing a change across, say, 20 articles, by just scrolling down vs having to make 20 selections then scroll. Just counted - it's up to 22 articles now. Per our earlier chat I am open to switching this when we refactor article selector to be a more generic demo control - as I mentioned I can include an 'all' option to preserve the accelerated proofing workflow at that time.

demo/ThemeTransform.html
montehurd: Array.from() wasn't needed. Good catch! Fixed!

querySelectorAll is needed though.

montehurd: This method was incorrectly named though. Fixed and simplified it a bit.

demo/ThemeTransform.html
montehurd: It's flattening an array of arrays. There's now a named method for this.

demo/ThemeTransform.html
montehurd: This is no longer in this file.

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Model for article data in ./article.json
+ * @type {object}
+ */
+class ArticleRef {
+  /**
+   * Constructor
+   * @param {!string} lang
+   * @param {!string} title
+   * @param {number} revision
```
niedzielski: Please add nullity (either `?` or `!`).

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Model for article data in ./article.json
+ * @type {object}
```
niedzielski: I don't think a type is necessary here because the type is ArticleRef. If you want, you could add `@class` but I'm not sure why that's needed since the class keyword is used.

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Model for article data in ./article.json
```
niedzielski: Very minor typo: this filename ended up being pluralized as articles.json.

demo/DemoArticles/ArticleRef.js
niedzielski: Whoops! I forgot this is `require`'d not `import`'d.

demo/DemoArticles/articles.json
niedzielski: You'd get some nice typing at the source of creation. Instead of an anonymous object that you convert to a type, you just instantiate the type directly.

demo/DemoArticles/ArticleRef.js
niedzielski: Can you do `module.exports = ArticleRef` though (and update your `require`)?

demo/DemoArticles/update.js
```diff
@@ -0,0 +1,51 @@
...
+
+const fetchAndSaveJSONForArticleRef = articleRef => {
+  const requestOptions = {
+    method: 'POST',
+    uri: articleRef.url(),
+    encoding: 'utf-8',
+    gzip: true
+  }
+  const responseHandler = (error, response, body) => {
+    if (!error && response.statusCode === 200) {
```
niedzielski: Consider throwing on error or unexpected response to be consistent with error handling of `fs.writeFile()`.

demo/ThemeTransform.html
```diff
@@ -0,0 +1,203 @@
...
+  <body>
+    <div id='theme_demo_tools'>
+      <control-panel></control-panel>
+      <div id='theme_demo_menu'></div>
+    </div>
+    <div class='content' id='content'></div>
+    
+    <script>
+
+/**
```
niedzielski: The indentation for this appears to be incorrect. Any idea why the linter isn't picking this up?

rollup.config.js
```diff
@@ -2,21 +2,68 @@ import babel from 'rollup-plugin-babel'
 import css from 'rollup-plugin-css-porter'
 import pkg from './package.json'
 
+const fs = require('fs')
+
+/**
+ * Custom rollup plug-in which bundles all transforms' CSS in a single file.
+ * @param  {callback} options Rollup plug-in options func
```
niedzielski: Is the type actually `callback`? Should this be `?function` and the comment mention callback?

rollup.config.js
```diff
@@ -2,21 +2,68 @@ import babel from 'rollup-plugin-babel'
...
+}
+
 export default {
   entry: 'src/index.js',
   dest: pkg.main,
   format: 'umd',
   moduleName: 'pagelib',
   sourceMap: true,
   plugins: [
-    css({
```
niedzielski: This plugin should be removed from the package too.

rollup.config.js
```diff
@@ -2,21 +2,68 @@ import babel from 'rollup-plugin-babel'
...
 import pkg from './package.json'
 
+const fs = require('fs')
+
+/**
+ * Custom rollup plug-in which bundles all transforms' CSS in a single file.
+ * @param  {callback} options Rollup plug-in options func
+ * @return {void}
+ */
+const wikimediaPageLibraryTransformCSSRollupPlugin = (options = {}) => {
```
niedzielski: It's cool that you were able to implement this in so few lines of code (and very instructive!)! Since this is a little unusual for the library and something we'll have to maintain, I'd like to keep a short decision record as I'm sure we'll have to explain it to someone at some point. Have you considered the following alternatives?

- Importing (or requiring) ThemeTransform.css in index.js before all other imports. I actually wouldn't mind moving all CSS imports to index.js since order matters.
- Moving ThemeTransform.css to live in wikimedia-page-library-override.css (right now it's just called Empty.css) and adding a note to the readme that this is expected to be linked prior to the transform CSS. These overrides are theme specific but they're also quite general and this would give us the opportunity to break these up across multiple files like Math.css, Infobox.css, FancyExtension.css, and so forth.
- If it's inserted alphabetically, renaming ThemeTransform.css something like _ThemeTransform.css.
- Generating a new CSS file for theming
- Switching to another plugin or bundler that supports this

Please consider moving this plugin to a separate file in our library and filing a PR or ticket against the old plugin's repo.

rollup.config.js
```diff
@@ -2,21 +2,68 @@ import babel from 'rollup-plugin-babel'
...
+    },
+    onwrite(opts) {
+      if (!Object.keys(cssBuffer).length) { return }
+
+      const keys = Object.keys(cssBuffer)
+
+      // We want the theme transform to be first. This is because the theme transform CSS has to use
+      // some '!important' CSS modifiers to reliably set themes on elements which may contain inline
+      // styles. Moving it to the top of the file is necessary so other transforms can override
+      // these '!important' themes transform CSS bits if needed.
```
niedzielski: It's a bummer that this complexity has to exist within our own library. The usage of `!important` seems to always end in an arms race that concludes with `!important` inline style or JavaScript. I'm sorry you had to deal with this complexity but I'm glad I was able to learn from it (I certainly wasn't thinking about this when we started) :+1:

src/transform/CollapseTable.css
```diff
@@ -80,7 +80,7 @@ table.infobox caption {
 }
 
 /* The outermost table container. todo: rename. */
-.app_table_container {
+div.app_table_container {
```
niedzielski: I thought this class was specific to our library and only ever applied to a div in the CollapseTableTransform?

src/transform/CollapseTable.css
```diff
@@ -89,14 +89,14 @@ table.infobox caption {
   clear: both;
 }
 
-.pagelib-theme-dark .app_table_container {
-  box-shadow: none;
-  background-color: #202020;
+.pagelib-theme-dark div.app_table_container {
+  box-shadow: none !important;
```
niedzielski: Kind of the same question as above. I thought this was the pagelib's class. Who are we overriding with `!important`?

src/transform/ThemeTransform.css
niedzielski: :+1: Thanks!

demo/DemoArticles/articles.json
montehurd: Oh god I misunderstood your initial comment! :) I get it now!

montehurd:
@niedzielski Good news! I got the theme demo update script to fetch MCS as well as MobileView, and the demo file to show them!

montehurd:
@niedzielski Better news - so far no issues specific to MCS on dark or sepia themes!

montehurd:
Will tackle other CR comments tomorrow :)

rollup.config.js
montehurd: Ya if Wikipedia forbade inline styles we woundn't have to do this :(

demo/DemoArticles/articles.json
niedzielski: Now I feel dumb about this. Maybe we should just assume that every article is supported on _both_ MediaWiki action API and the Content Service backends rather doubling the data.

demo/DemoArticles/update.js
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env node
+/* eslint-disable require-jsdoc */
+
+const fs = require('fs')
+const request = require('request')
```
niedzielski: This comment isn't intended to sway you either way but here's an example Jon Robson did using node-fetch: https://gist.github.com/jdlrobson/a358f9ed079fbc9ebf18

rollup.config.js
montehurd: > Importing (or requiring) ThemeTransform.css in index.js before all other imports. I actually wouldn't mind moving all CSS imports to index.js since order matters.

I like this much better... switching and I'll remove the custom rollup plugin.

montehurd: Or maybe I'll simplify it but keep it instead of using the 'porter'... hmm...

montehurd: Ok was able to really simplify this. All I had to do was add a `sort-imports` eslint exception to index.js

I also was able to vastly simplify the custom rollup plugin and just use it instead of the `porter` plugin for both css files.

src/transform/CollapseTable.css
montehurd: This is fallout from the theme-ing 'baseline' approach. Had to use more aggressive CSS to robustly apply themes to un-classed elements with inline styles. Transforms must match or exceed the theme level of selector specificity to be able to override the 'baseline' styles set by the theme transform.

src/transform/CollapseTable.css
montehurd: Same as [above](https://github.com/wikimedia/wikimedia-page-library/pull/77#discussion_r131987125).

rollup.config.js
montehurd: Fixed!

rollup.config.js
montehurd: Fixed!

demo/ThemeTransform.html
montehurd: Hmm not sure...

demo/DemoArticles/update.js
montehurd: Fixed!

demo/DemoArticles/ArticleRef.js
montehurd: Fixed!

demo/DemoArticles/ArticleRef.js
montehurd: Fixed!

demo/DemoArticles/ArticleRef.js
montehurd: Fixed!

demo/DemoArticles/articles.json
montehurd: Yes I came to the same conclusion independently and implemented it like that :)

demo/DemoArticles/update.js
montehurd: Ok to roll with 'request' for now and maybe tweak later?

src/transform/ThemeTransform.css
montehurd: > may be just as likely to break elements that were intentionally styled differently

I think it's worth keeping the theme user story in mind - which is precisely about "overriding elements which were intentionally styled differently". The '[baseline](https://github.com/wikimedia/wikimedia-page-library/pull/77#discussion_r131254971)' approach is unfortunately the only way I know to robustly apply themes to inline styles. I've not seen any "break(ing) elements" with this approach, just the disappearance of all known issues :)

-----

Oh, side note - I just updated all of the `after` screenshots.

One more - I also added articles from OTRS article theming bug reports to the theme demo article list and confirmed reported issues are resolved with the PR.

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,108 @@
+
+/**
+ * Type representing article source endpoints.
+ * @enum {ArticleRefSourceType}
+ */
+const ArticleRefSourceType = {
+  mobileView: 1,
```
niedzielski: I know this is an enum pattern but I think you'll find your filenames more readable if you use a string here or convert from enum to string fileName().

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,108 @@
...
+    this.title = title
+    this.revision = revision
+    this.sourceType = sourceType
+  }
+
+  /**
+   * String which can be used to for serialization.
+   * @return {!string}
+   */
+  fileName() {
```
niedzielski: I believe N should be lowercase but obviously not a show-stopper.

https://en.wikipedia.org/wiki/Filename

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,108 @@
...
+    return `${this.lang} > ${this.title} > ${displayNameForArticleRefSourceType(this.sourceType)}`
+  }
+
+  /**
+   * URL string used to fetch JSON for this article revision.
+   * @return {!string}
+   */
+  url() {
+    switch (this.sourceType) {
+    case ArticleRefSourceType.mobileView:
```
niedzielski: These switches would be better encapsulated as properties on the enum itself but it's fine to punt on this for now.

demo/DemoArticles/ArticleRef.js
```diff
@@ -0,0 +1,108 @@
...
+      return [json.lead.sections[0]].concat(json.remaining.sections)
+    default:
+      return []
+    }
+  }
+
+  /**
+   * Fetch promise resolving to array of section JSON from local data file.
+   * @param {!string} dataFilePath
+   * @return {!Promise}
```
niedzielski: For future reference, please parameterize your Promises so we know what they resolve to!

https://github.com/jsdoc3/jsdoc/issues/1197#issuecomment-312948746

demo/DemoArticles/update.js
niedzielski: :+1: Sure, comment was just another reference for reading.

demo/DemoArticles/update.js
```diff
@@ -0,0 +1,77 @@
...
+
+const DATA_PATH = './data/'
+
+const articleRefs = flattenArrayOfArrays(
+  indexJSON.map(articleData => [
+    new ArticleRef(
+      articleData.lang,
+      articleData.title,
+      articleData.revision,
+      ArticleRefSourceType.mobileView
```
niedzielski: For future refactor consideration: storing the backend type is needlessly a property of ArticleRef. The client of ArticleRef can just supply this to a `ArticleRef.url(ArticleRefSourceType)` method or (even better) supply the ArticleRef to `ArticleRefSourceType.url(ArticleRef)`.

demo/DemoArticles/update.js
```diff
@@ -0,0 +1,77 @@
...
+      articleData.revision,
+      ArticleRefSourceType.mobileContentService
+    )
+  ])
+)
+
+// eslint-disable-next-line no-console
+console.log(`Fetching JSON for ${articleRefs.length} titles...`)
+
+// Linting complained of some improperly formatted characters which turned out to be these.
```
niedzielski: todo: try to document what's happening here better, implement this generically instead of case by case, and consider pushing this fix to the server (since they return JSON responses).

demo/ThemeTransform.html
```diff
@@ -0,0 +1,185 @@
...
+  <body>
+    <div id='theme_demo_tools'>
+      <control-panel></control-panel>
+      <div id='theme_demo_menu'></div>
+    </div>
+    <div class='content' id='content'></div>
+    
+    <script>
+
+/* global ArticleRef, ArticleRefSourceType, flattenArrayOfArrays */
```
niedzielski: Consider indenting this whole script even though are linter isn't working.

rollup.cssBundler.js
```diff
@@ -0,0 +1,74 @@
+const fs = require('fs')
```
niedzielski: I personally would prefer one of the alternatives I mentioned to maintaining our own custom Rollup plugin but this is fine for merging (and a neat reference).

changelog.md
```diff
@@ -1,3 +1,10 @@
+### v4.5.5
```
niedzielski: Please bump version minor for new release

demo/DemoArticles/update.js
montehurd: > The client of ArticleRef can just supply this...

True, but there's a big benefit to more complete ArticleRef modeling - Client code only has to specify the sourceType _once_ when creating an ArticleRef, and later calls to its methods which are variant on sourceType (sections, url, filename, display name, etc) _just work_ without having to maintain external sourceType state. 

This made it much simpler to use them (via functional fun) in multiple places in very different ways - i.e. by the update script to fetch, by the demo to construct menu items and also to present sections.

demo/DemoArticles/update.js
montehurd: I honestly have no idea what's going on with these characters. Also unsure how this could be generalized...

rollup.cssBundler.js
montehurd: I implemented the [first part of your first alternative](https://github.com/wikimedia/wikimedia-page-library/pull/77#discussion_r131291184):

> Importing (or requiring) ThemeTransform.css in index.js before all other imports.

I just also kept the custom rollup plugin because it meant we could add comments for each section of CSS (in the resulting file) saying which transform it originated from.

montehurd: > I just also kept the custom rollup plugin because it meant we could add comments for each section of CSS (in the resulting file) saying which transform it originated from.

^ not sure that's as useful as I thought it was at the time though... :)

montehurd: iirc Rollup passes plugins file references in the order they were included. (We include transforms in `src/transform/index.js` and those transforms include any CSS they use.) The old plug-in kept an internal list of these file references which matched a patterns - i.e. the CSS ones. 

The problem was it did something with this list which re-ordered it. The custom plugin preserves the order. 

I did have to move the theme transform to be [first](https://github.com/wikimedia/wikimedia-page-library/pull/77/commits/4027cb89e392f5193a055f87b73c4e972eabcbf7#diff-9dd9e543fe2f7d30f9fe16b64e769564) in `src/transform/index.js`.

demo/DemoArticles/update.js
niedzielski: Did you see my link? json.org specifies that a JSON string is composed of chars. chars are:
- any-Unicode-character-except-"-or-\-or-control-character
- \"
- \\
- \/
- \b
- \f
- \n
- \r
- \t
- \u four-hex-digits

This code is just escaping it it in the dumbest way possible. We just need to generically map all illegal characters to escaped versions of those characters and decide whether to escape them as Unicode characters (current implementation) or HTML entities.

It could also just be that our JSON linter isn't working right but I think the first order of investigation would be to peak at the linter source (especially since it's likely written in JavaScript).

demo/DemoArticles/ArticleRef.js
montehurd: Fixed!

demo/DemoArticles/update.js
niedzielski: I think I'll have to think on it longer to appreciate your statement fully. Please however consider that ArticleRefSourceType can easily become outdated when a new type is added and new states aren't added to _every_ place it's referenced. Stateless polymorphism avoids conditionals.

changelog.md
montehurd: But isn't this a patch?

niedzielski: Oops! I just read the "New" prefix which means minor bump. That should be "Chore" since it's not a functional change.

demo/DemoArticles/update.js
montehurd: > Please however consider that ArticleRefSourceType can easily become outdated when a new type is added and new states aren't added to every place it's referenced.

I *think* it's still easier to add them this way because all of the client logic is type agnostic (with exception of creation of new ArticleRef's).

niedzielski:
:1st_place_medal: Published as NPM package v4.5.5. Please update your integration patch to reference this version (and don't forget to unlink before building the bundle).

niedzielski:
+@mdholloway FYI :]

PR #78, opened 2017-08-15
Hide edit pencil on iOS on H3, H4, H5 and H6 on iOS.

montehurd:
https://phabricator.wikimedia.org/T171355#3495253

### DIFF
--- a/src/transform/EditTransform.css
+++ b/src/transform/EditTransform.css
@@ -86,5 +86,5 @@ a.pagelib_edit_section_link {
 .pagelib-platform-ios h4.section_heading span.pagelib_edit_section_link_container,
 .pagelib-platform-ios h5.section_heading span.pagelib_edit_section_link_container,
 .pagelib-platform-ios h6.section_heading span.pagelib_edit_section_link_container {
-    display: none !important;
+    display: none;
 }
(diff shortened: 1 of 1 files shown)


src/transform/EditTransform.css
```diff
@@ -80,4 +80,11 @@ a.pagelib_edit_section_link {
...
   background-color: #111;
-}
\ No newline at end of file
+}
+
+.pagelib-platform-ios h3.section_heading span.pagelib_edit_section_link_container,
+.pagelib-platform-ios h4.section_heading span.pagelib_edit_section_link_container,
+.pagelib-platform-ios h5.section_heading span.pagelib_edit_section_link_container,
+.pagelib-platform-ios h6.section_heading span.pagelib_edit_section_link_container {
+    display: none !important;
```
niedzielski: Sorry but it isn't clear why `important` priority is needed here and I didn't see any comments in the commit, the PR, or the ticket. Since we add these edit pencils ourselves, I'm hoping we can avoid specifying `important` priority. (This isn't just me hating on `important`, BTW, it's even in the [MediaWiki CSS recommendations](https://www.mediawiki.org/wiki/Manual:Coding_conventions/CSS#.21important)!)

montehurd: Oops fixed!

montehurd: Agreed with your reasoning. Only really needed it with the theme changes for the inline style issue.

PR #79, opened 2017-08-15
Fix querySelectorAll polyfill return value for null elements.

montehurd:


src/transform/Polyfill.js
```diff
@@ -22,8 +22,12 @@ const matchesSelector = (el, selector) => {
  * @param {!string} selector
```
niedzielski: The JSDoc says nonnull element. Regardless, this feels like it addresses the symptom instead of the cause. What is this PR for? Clients should never call `element.querySelectorAll()` or `element.foo` if element itself is null. I would prefer we avoid defensively handling every null scenario possible in every method as it makes the code difficult to reason about.

montehurd: Ah that's a good point! Closing.

PR #80, opened 2017-08-15
Remove line breaks from selector string.

montehurd:


src/transform/FooterContainer.js
```diff
@@ -22,11 +22,12 @@ const updateBottomPaddingToAllowReadMoreToScrollToTop = window => {
  * @return {void}
  */
 const updateLeftAndRightMargin = (margin, document) => {
-  const elements = Polyfill.querySelectorAll(document, `
-    #pagelib_footer_container_menu_heading,
-    #pagelib_footer_container_readmore,
-    #pagelib_footer_container_legal
-  `)
+  const selectors = [
```
niedzielski: Is this fixing some weird platform bug or is it a purely nonfunctional clean-up change / chore?

montehurd: In the js file produced by babel I noticed it had '\n' inside of this string and it just seemed odd. Didn't appear to affect functionality, but thought it was easy to remove those.

niedzielski: Find with me. Just wanted to double check I understood the change!

montehurd: Sorry I should have had a better description! :)

PR #81, opened 2017-08-15
Doc: update changelog

montehurd:


changelog.md
```diff
@@ -1,3 +1,6 @@
+### v4.6.0
+- New: hide edit pencils on iOS on H3, H4, H5 and H6
```
niedzielski: Do you consider this a new feature? I think this would be either an update or a fix but I'll leave it to you. I normally use the ESLint project as an example when I'm thinking of release notes. Here's a [smorgasbord example](https://github.com/eslint/eslint/releases/tag/v4.0.0-alpha.0) from one of their releases.

montehurd: I was unsure. I'm cool with switching it to a `fix` and patch. Updating...

montehurd: Fixed!

PR #82, opened 2017-08-16
Read more text line height with smallest-size system font is too low

nambatee:
https://phabricator.wikimedia.org/T169262

Before

before

After

after

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,8 @@
+### v4.5.7
+- Fix: add a few css properties to read more part of the footer
+
 ### v4.5.6
 - Fix: hide edit pencils on iOS on H3, H4, H5 and H6
-- Fix: add a few css properties to read more part of the footer
 
 ### v4.5.5
 - Fix: dark and sepia themes fixes for various elements including those having inline style colors
--- a/changelog.md
+++ b/changelog.md
@@ -1,8 +1,6 @@
 ### v4.5.6
 - Fix: hide edit pencils on iOS on H3, H4, H5 and H6
-
-### v4.5.6
-- Fix: add a few css properties to read more part of the footer 
+- Fix: add a few css properties to read more part of the footer
 
 ### v4.5.5
 - Fix: dark and sepia themes fixes for various elements including those having inline style colors
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,9 @@
 ### v4.5.6
 - Fix: hide edit pencils on iOS on H3, H4, H5 and H6
 
+### v4.5.6
+- Fix: add a few css properties to read more part of the footer 
+
 ### v4.5.5
 - Fix: dark and sepia themes fixes for various elements including those having inline style colors
 - Chore: add theme demo. To add articles to theme demo update `articles.json` then run `update.js`
--- a/src/transform/FooterContainer.css
+++ b/src/transform/FooterContainer.css
@@ -28,7 +28,7 @@
   display: table;
   width: 100%;
   border-collapse: collapse;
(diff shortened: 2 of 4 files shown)


changelog.md
```diff
@@ -1,5 +1,6 @@
 ### v4.5.6
 - Fix: hide edit pencils on iOS on H3, H4, H5 and H6
+- Fix: add a few css properties to read more part of the footer
```
niedzielski: I know how we have this is a bit odd. These release notes are backwards facing the moment a release is cut and future facing before the next release. Since v4.5.6 has been published, we can't make changes to it. As this is a bug fix, please add a new v4.5.7 section. :+1:

berndsi: I recommend having a placeholder heading for not yet released changes, similar to what [domino](https://github.com/fgnass/domino/commit/e43984f7c80023943b6585652be72aeafb4e2971) does.

src/transform/FooterContainer.css
```diff
@@ -28,6 +28,7 @@
   display: table;
   width: 100%;
   border-collapse: collapse;
+  line-height:22px;
```
berndsi: This is a more general comment. Sizes like these are probably better specified in em (or even better in rem).

montehurd: I think this is correct. @nambatee want to give this a try?

montehurd:
@nambatee 

Cool I'll check out your changes in the morning. 

Oh one thing to consider tweaking in the future is making your PR titles better describe the specific bug which was fixed.

montehurd:
Also, on device I'm seeing other unwanted additional space:

# Before
screen shot 2017-08-21 at 11 10 20 am

# After
screen shot 2017-08-21 at 10 58 09 am

nambatee:
@mhurd the space beneath the save button in my `after` picture is there because the cell is expanded to fit the thumbnail. Is that the space that you're talking about or is there something I'm missing?

As for the additional unwanted space, it's caused by the previously suggested px -> em change.

montehurd:
@nambatee 

> Since em is relative to font size and in this case we're manipulating the font size, I don't see how using em instead of px is solving the problem in this case.

I see. I think we may have to make more invasive changes to the structure of the read more cells to make them not go to wonky town at small text sizes...

If I get my on-this-day endpoint changes wrapped today I can take a deeper dive tomorrow and hopefully make more useful suggestions :)

nambatee:
Is using `px` instead of `em` an option? The original solution solves it, without invasive changes.

montehurd:
@nambatee Ah good to know - I'll re-test that commit...

nambatee:
`px`:

screen shot 2017-08-21 at 7 15 04 pm

`em`:

screen shot 2017-08-21 at 7 16 02 pm

For regular font size.

montehurd:
px still has a space issue...

29541900-231b3bb2-86a5-11e7-8ccc-807a0f5a357c

montehurd:
or is that not appearing on device?

nambatee:
It is @montehurd. My bad, for the previous screen I didn't revert `min-height` to be using `px` instead of `em`.

This is what the original commit looked like - 

screen shot 2017-08-21 at 7 23 43 pm

montehurd:
@nambatee oh which min-height? if it's just the one for the overall cell you could just tweak it until it's the same height (for default text size), then test that things hold together when you increase or decrease text size...

nambatee:
@montehurd `min-height` for `pagelib_footer_readmore_page`, one of the 3 properties changed in the PR.

PR #83, opened 2017-08-16
Fix dark and sepia theming on articles which contain color swatches and sports uniforms

montehurd:
https://phabricator.wikimedia.org/T173449

# enwiki > Pantone

## Before
screen shot 2017-08-16 at 4 35 02 pm

## After
screen shot 2017-08-16 at 4 32 58 pm

# enwiki > Manchester United F.C.

## Before
screen shot 2017-08-18 at 9 03 53 pm

## After
screen shot 2017-08-18 at 9 01 28 pm


src/transform/ThemeTransform.css
```diff
@@ -46,15 +46,16 @@ styles with only the things which need tweaking.
 }
 
 /* baseline div, span, ul, li */
-.pagelib-theme-dark .content div,
+/* en > Pantone > 792312384 */
+.pagelib-theme-dark .content :not(.color_swatch) > div,
```
niedzielski: I don't mind merging this for a quick fix since you've tested so many articles for regressions but I think this problem would be better solved and less fragile with the classification techniques used in ThemeTransform.classifyElements(). There's some discussion on the trouble with not-ing ancestors [here](https://stackoverflow.com/a/7084147/970346).

montehurd: Oh cool thanks I'll take a look at the classify bits!

montehurd:
@niedzielski The commits I just added changed things around a bit. Probably needs a re-review :)

src/transform/ThemeTransform.js
```diff
@@ -2,11 +2,11 @@ import './ThemeTransform.css'
 import ElementUtilities from './ElementUtilities'
 import Polyfill from './Polyfill'
 
-// Elements marked with either of these classes indicate certain ancestry constraints that are
+// Elements marked with these classes indicate certain ancestry constraints that are
 // difficult to describe as CSS selectors.
 const CONSTRAINT = {
   IMAGE_NO_BACKGROUND: 'pagelib-theme-image-no-background',
-  IMAGE_NONTABULAR: 'pagelib-theme-image-nontabular'
+  DIV_DO_NOT_APPLY_BASELINE: 'pagelib-theme-div-do-not-apply-baseline'
```
niedzielski: What do you think about using that underscore naming convention we keep talking about? I know it'd look funky in contrast with the old style but it is the way forward.

Also, consider naming the style for what its classifying, not what we want to do with it at the moment. For example, if we had a classifier that identified large houses and small houses, we could call the rules `pagelib_theme_large_house` and `pagelib_theme_small_house` and then choose to style them red and blue. Because we know we want them to be styled as red and blue, we could have also called them `pagelib_theme_red_house` and `pagelib_theme_blue_house` but it'll be brittle if we ever want a new paint job. If a small house ever becomes a large house, it will be classified as `pagelib_theme_large_house` which is what we want.

Another example that's confusing is identifying `IMAGE_NO_BACKGROUND` images with a function called `imageNeedsWhiteBackground()`. Images classified as "no background" normally _do_ need a white background but not always. When we say in the code that they need a white background, we're not consider all the inputs AND we're putting styling concerns in JavaScript. Classification is supposed to be very pointed and each class is supposed to identify one thing about an element that's difficult to describe in pure CSS. For example, there is a already a case where an image that "needs a white background" should _not_ have a white background because there are other inputs (math formulas). You can see how this was a similarly confusing case in the original Android implementation [here](https://phabricator.wikimedia.org/diffusion/APAW/browse/master/www/js/dark.js;ccc713d7abdbc9261f9ee55a4311d7645eca1783$41). `classifyElements()` is just annotating elements for what we can assert about them in JavaScript and should have no knowledge of styling.

src/transform/ThemeTransform.js
```diff
@@ -41,14 +62,18 @@ const setTheme = (document, theme) => {
...
+    .filter(imageNeedsWhiteBackground)
+    .forEach(image => {
       image.classList.add(CONSTRAINT.IMAGE_NO_BACKGROUND)
-    }
-    if (!ElementUtilities.isNestedInTable(image)) {
-      image.classList.add(CONSTRAINT.IMAGE_NONTABULAR)
-    }
-  })
+    })
+  /* en > Away colours > 793128975 */
```
niedzielski: We generally use line comments within functions.

src/transform/ThemeTransform.js
```diff
@@ -41,14 +62,18 @@ const setTheme = (document, theme) => {
...
+  /* en > Manchester United F.C. > 793244653 */
+  /* en > Pantone > 792312384 */
+  Polyfill.querySelectorAll(element, 'div.color_swatch div, div[style*="position: absolute"]')
```
niedzielski: I don't think I would have taken this approach to classify elements that we _don't_ want styled. After some consideration, I like it and believe that identifying exceptions to generally be superior to classifying everything else instead. The reason is that, in general, if I was writing a given article, I'd want the whole thing to be styled by default and to manually, where needed, what shouldn't be. The only unfortunate side-effect is that we have to use `:not()` but this is already how math formulas are handled. Good work!

src/transform/ThemeTransform.js
```diff
@@ -33,6 +33,27 @@ const setTheme = (document, theme) => {
   }
 }
 
+/* en > Away colours > 793128975 */
```
niedzielski: Consider moving these examples into the JSDocs.

src/transform/ThemeTransform.js
niedzielski: Another similar example is saying "bookmark_filled" to refer to the read more saved page icon instead of just generically "icon_enabled". What if the icon is changed to a diskette? What should just be an asset swap now becomes a project-wide scouring, possibly API breaking change.

Specificity can help in conceptually scoping down a problem to focus on writing an initial concrete version but must be immediately revised (before pushing!) to remove needless detail. The reasons are that:
- We lose a ton of flexibility to do even the most subsequent basic changes or refactoring until the code is reconsidered, reimagined, and rewritten to be malleable.
- The code reads as a one-off, custom-made implementation which discourages consideration for nearly identical problems with different details. If we make a function that sorts hot dogs but can just as easily sort cylinders, we probably shouldn't call it `sortHotDogs()`.
- The code quickly becomes outdated because many details, relevant or not, are captured and any change in the problem requires resynchronization with the code whether it was functionally important or not.
- In some cases, it's actually more difficult to read because it isn't clear why it's so specific. It'd be like asking for directions to the grocery store and someone describing every turn but also what kind of paint was used for the lines on each road. The detail is only distracting noise! Just tell me what turns to take!

An equivalent example would be needlessly naming a variable `fooList` instead of `foos` or `quicksort` instead of `sort`. When we change the list to an array, or the quicksort to a bubble sort, instead of just changing the implementation strategy (hopefully one line) we have to change _all_ references and then _struggle_ to understand if there was a specific characteristic of the lists or quicksort that was really needed or this just a case of needless specificity and there were no assumptions on the underlying collection or algorithm type. There may be some but the burden of identifying important characteristics and removing unimportant detail should be a one-time cost paid by the author, not every reader.

This all boils down to writing code that _only_ references the behavior / characteristic / interface we **need** instead of today's happenstance implementation / strategy. Another way to put it is to try to only describe what's so important, so fundamental, we can't imagine or execute our code snippet without it.

src/transform/ThemeTransform.js
montehurd: Ya I struggled with this - there were issues with trying to get `:not()` to work with multiple exceptions. Didn't seems to have widespread support.

montehurd:
@niedzielski thanks for the CR! i'll digest it tomorrow :)

src/transform/ThemeTransform.js
```diff
@@ -33,6 +33,27 @@ const setTheme = (document, theme) => {
...
+/* en > Manchester United F.C. > 793244653 */
+/**
+ * Determines whether white background should be added to image.
+ * @param  {!HTMLImageElement} image
+ * @return {!boolean}
+ */
+const imageNeedsWhiteBackground = image => {
+  const src = image.src
+  if (src.endsWith('.svg.png')) {
+    return !(
```
montehurd: It's broadly used by a class of soccer articles.

test/transform/ThemeTransform.test.js
```diff
@@ -49,21 +49,5 @@ describe('ThemeTransform', () => {
       const clazz = pagelib.ThemeTransform.CONSTRAINT.IMAGE_NO_BACKGROUND
       assert.ok(document.querySelector('img').classList.contains(clazz))
     })
-
```
montehurd: I was able to remove the tabular/non-tabular determination, so the test was no longer meaningful. If we can get things configured so we can add a couple of basic visual tests I'd like to circle back and add tests for each of these tricky theming areas - it's getting unwieldy even with the help of the theming demo to test and experiment with revisions in a safe way.

src/transform/ThemeTransform.js
montehurd: Was also fallout from not being able to use compound `:not()`s.

src/transform/ThemeTransform.js
montehurd: This is all really good information, but the reason I needed to switch to a `pagelib-theme-div-do-not-apply-baseline` approach is CSS doesn't seem to handle multiple `:not()`'s. If you look at the theming CSS for images specifically it was already very difficult to reason about, and w/o being able to specify additional `:not()` classes it couldn't be extended.

src/transform/ThemeTransform.js
montehurd: Oh I was using the same convention I used in the theme CSS file for consistency when referencing articles where issues occurred.

montehurd:
@niedzielski I think I addressed all of the issues we'd identified.

PR #87, opened 2017-08-24
Add collapse tables checkbox to theme demo.

montehurd:
I took a stab at adding a collapse tables checkbox to ControlPanel.html, but ran into some issues. Rather than going down that rabbit hole (since we may refactor this to use iFrame anyway) I just added a quick checkbox to the theme demo so at least I can proof theme changes against table collapsing in mean time.

### DIFF
--- a/demo/ThemeTransform.html
+++ b/demo/ThemeTransform.html
@@ -51,6 +51,7 @@
   <body>
     <div id='theme_demo_tools'>
       <control-panel></control-panel>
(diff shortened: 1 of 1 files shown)


demo/ThemeTransform.html
```diff
@@ -51,6 +51,7 @@
   <body>
     <div id='theme_demo_tools'>
       <control-panel></control-panel>
+      <input type='checkbox' id='collapse_tables_checkbox'> <label for=collapse_tables_checkbox>Collapse tables</label>
```
niedzielski: A checkbox doesn't seem quite right here since there's no undo-ing the transform but I think this is a known issue.

PR #89, opened 2017-09-03
Use 'pagelib_' prefix and snake_case consistently for transform CSS class naming

montehurd:


niedzielski:
:+1: thanks for this change! Since so many lines changed, please do cut a release with just this in it so it's easy for Android to verify no regressions occurred.

montehurd:
> Since so many lines changed, please do cut a release with just this in it so it's easy for Android to verify no regressions occurred.

Will do!

montehurd:
@niedzielski done!

PR #90, opened 2017-09-21
Add link out to an article's Talk page as an item on the footer menu

nambatee:
https://phabricator.wikimedia.org/T172747

Anyone knows how we generate data URIs for footer menu icons?


berndsi:
I think @nambatee's question is still a good one. I'd like to know, too, how you generate the embedded data urls in your repo. In the MobileApp extension it is done using the [Embed styling](https://www.mediawiki.org/wiki/ResourceLoader/Features#Resource:_Styles) (see /* embed */ comment in [less files](https://phabricator.wikimedia.org/diffusion/EMOB/browse/master/styles/android/ipa.less;1a7d6c3a183ad2c3b4a2585fd93952e14306cf96$4)).

In this repo I only see css files, though.

niedzielski:
Sorry, I missed the note in the review! We currently only support CSS in the repo so the data URIs are currently generated manually. We'd like to add support for generating the URIs from source images but don't currently have that functionality yet.

montehurd:
@niedzielski 
> We'd like to add support for generating the URIs from source images but don't currently have that functionality yet.

Now that I've mucked about with [custom rollup plug-ins](https://github.com/wikimedia/wikimedia-page-library/blob/master/rollup.cssBundler.js) I may be bold enough to attempt having the build product also deliver an `images` directory containing actual image assets (may try a spike...), but we'll have to think about how this complicates integration a bit...

montehurd:
@nambatee 

> Anyone knows how we generate data URIs for footer menu icons?

iirc I used https://www.base64-image.de - but it looks like whatever you used produced similar output. 

I'll do a quick integration test & review, but so far looks solid :)

niedzielski:
@montehurd, I also wouldn't mind just switching to Webpack for all the plugins and configurability it offers. We're using this in Marvin and after some initial investment, it's been great. If you pursue a rollup.js option, please do try to an existing solution before writing your own! We probably don't want to maintain plugins in this repo if it can be avoided.

montehurd:
@nambatee 
Oh would you mind updating the changelog? Then if all goes well with testing I can just merge it.

montehurd:
@niedzielski 
> I also wouldn't mind just switching to Webpack for all the plugins and configurability it offers.

I am 100% down with that if you recommend it :) Would much rather not have to roll our own anymore.

PR #92, opened 2017-09-21
Add "View in browser" as a footer link

nambatee:
https://phabricator.wikimedia.org/T172746

simulator screen shot - iphone 6 plus - 2017-09-22 at 01 29 07

Suggestions how to make it better more than welcome 😁

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+### v4.7.0
+- New: add a footer link to view the article in browser
+
 ### v4.6.0
 - New: add a link to a talk page as a footer menu item
 
--- a/src/transform/FooterLegal.js
+++ b/src/transform/FooterLegal.js
@@ -17,7 +17,7 @@ import './FooterLegal.css'
  * @param {?string} licenseSubstitutionString
  * @param {!string} containerID
  * @param {!FooterLegalClickCallback} licenseLinkClickHandler
- * @param {?string} viewInBrowserString
+ * @param {!string} viewInBrowserString
  * @param {!FooterBrowserClickCallback} browserLinkClickHandler
  * @return {void}
  */
--- a/src/transform/FooterLegal.js
+++ b/src/transform/FooterLegal.js
@@ -6,16 +6,23 @@ import './FooterLegal.css'
  */
 
 /**
+  * @typedef {function} FooterBrowserClickCallback
+  * @return {void}
+  */
+
+/**
  * Adds legal footer html to 'containerID' element.
  * @param {!Element} content
  * @param {?string} licenseString
  * @param {?string} licenseSubstitutionString
  * @param {!string} containerID
  * @param {!FooterLegalClickCallback} licenseLinkClickHandler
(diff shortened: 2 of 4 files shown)


src/transform/FooterBrowserLink.css
```diff
@@ -12,6 +12,7 @@
   line-height: 1.5;
   display: block;
   margin-top: -10px;
+  padding-bottom: 27px;
```
nambatee: simulator screen shot - iphone 6 plus - 2017-09-22 at 01 50 02

changelog.md
```diff
@@ -1,3 +1,6 @@
+### v4.6.0
+- New: add a footer link to view the article in browser
```
montehurd: Oops sorry I published `4.6.0`, so we'll need to bump this to `4.7.0`

montehurd:
@nambatee 

Thoughts on maybe putting this in the `FooterLegal.js` transform? I didn't mean for that transform to only ever contain the footer legal sentence, but I guess I named it poorly :(

nambatee:
@montehurd I was thinking about at first that but you're right, given the name, I was a bit worried I'd be pushing non-legal stuff into the legal part of the footer :P I can consolidate it though, if we find a more generic name for that part of the footer

src/transform/FooterLegal.js
```diff
@@ -6,16 +6,23 @@ import './FooterLegal.css'
...
+  */
+
+/**
  * Adds legal footer html to 'containerID' element.
  * @param {!Element} content
  * @param {?string} licenseString
  * @param {?string} licenseSubstitutionString
  * @param {!string} containerID
  * @param {!FooterLegalClickCallback} licenseLinkClickHandler
+ * @param {?string} viewInBrowserString
```
niedzielski: Should this  be nonnull? When `viewInBrowserString` is `undefined`, `` `foo${viewInBrowserString}bar` === "fooundefinedbar"`` and the callback must be nonnull which is a little odd for an empty link.

src/transform/FooterLegal.js
```diff
@@ -29,13 +36,24 @@ const add = (content, licenseString, licenseSubstitutionString, containerID,
         ${licenseSubstitutionString}
       </a>
       ${licenseStringHalves[1]}
+      <br>
+      <div class="pagelib_footer_browser">
+      <a class='pagelib_footer_browser_link'>
```
niedzielski: 41-43 seem to be missing tag indentation which is inconsistent with the rest of this string..

nambatee: Did you mean indenting 42 (`${viewInBrowserString}`) or entire 41-43? 😃

niedzielski: Line 31 is a div and it's children are indented. I was thinking this code should be consistent with that styles since it's all one block. :+1:

nambatee:
Thanks @montehurd! No worries about the nom publish stuff, we can go through that when you're back :)

PR #93, opened 2017-10-12
Fix: override inline heights in Parsoid HTML

mdholloway:
Parsoid HTML image ancestors, unlike in MediaWiki HTML, often specify
height attributes.  These cramp widened images so that they overlap
their captions and one another (https://phabricator.wikimedia.org/T176753). 

Overriding with `height: 'auto'` fixes this issue.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+### v4.7.1
+- Fix: override image ancestor inline heights (found in Parsoid HTML) with `height: 'auto'`
+
 ### v4.7.0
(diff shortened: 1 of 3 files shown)


montehurd:
@mdholloway Oh interesting! Thanks for the PR! I'll do a quick iOS integration/tire-kicking, but may not get to it until tomorrow morning.

mdholloway:
Ping!  I'm hoping this fix can get into the next Android app release.  T176753 makes for a pretty crummy reading experience.

PR #94, opened 2017-10-26
Revert version bump.

montehurd:
'npm version patch' will do this. Wasn't sure if version would get double-bumped...


montehurd:
Self-merging this as it will get immediately re-set when I do 'npm version patch'.

PR #96, opened 2017-11-02
Fix graphs and charts issues on Sepia and Dark modes

nambatee:
https://phabricator.wikimedia.org/T179127

This is not the nicest fix but it does fix the graph & chart issues described in the ticket (on iOS). Better alternatives to tackle this are more than welcome.

### DIFF
--- a/src/transform/ThemeTransform.css
+++ b/src/transform/ThemeTransform.css
@@ -31,14 +31,14 @@ styles with only the things which need tweaking.
 
 /* baseline table */
 .pagelib_theme_dark .content table,
-.pagelib_theme_dark .content td,
+.pagelib_theme_dark .content td:not(.pagelib_theme_div_do_not_apply_baseline),
 .pagelib_theme_dark .content tr,
 .pagelib_theme_dark .content th {
   background: #27292d !important;
   color: #f8f9fa !important;
 }
 .pagelib_theme_sepia .content table,
-.pagelib_theme_sepia .content td,
+.pagelib_theme_sepia .content td:not(.pagelib_theme_div_do_not_apply_baseline),
 .pagelib_theme_sepia .content tr,
 .pagelib_theme_sepia .content th {
   background: #f0e6d6 !important;
@@ -53,7 +53,7 @@ styles with only the things which need tweaking.
 .pagelib_theme_dark .content li {
   background-color: inherit !important;
   color: #f8f9fa !important;
-  border-color: #cbc8c1 !important
+  border-color: #43464a !important;
 }
 .pagelib_theme_sepia .content div:not(.pagelib_theme_div_do_not_apply_baseline),
 .pagelib_theme_sepia .content span:not(.pagelib_theme_div_do_not_apply_baseline),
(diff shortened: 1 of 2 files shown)


montehurd:
Noticed a few regressions on the theme demo articles. At first glance I think the main issue is the query selectors which aren't scoped to `div.barbox` which causes `nots` to break baseline theme-ing on `td` and `span` elements which aren't in charts. Sorry I don't have more time to review/comment at the moment - was just heading out. Can pair tomorrow if you want.

src/transform/ThemeTransform.js
```diff
@@ -78,6 +78,41 @@ const classifyElements = element => {
     .forEach(div => {
       div.classList.add(CONSTRAINT.DIV_DO_NOT_APPLY_BASELINE)
```
niedzielski: What do you think about consolidating all these selector and just running this function once? Something like:

```js
const selector = [
  'div.color_swatch div',
  'div[style*="position: absolute"]',
  'div[role="img"] div',
  ...,
  ...,
  ...
].join()
Polyfill.querySelectorAll(element, selector).forEach(element =>
  element.classList.add(CONSTRAINT.DIV_DO_NOT_APPLY_BASELINE))
```

src/transform/ThemeTransform.css
```diff
@@ -94,7 +98,7 @@ styles with only the things which need tweaking.
 }
 
 /* caption color */
-/* en > Barack Obama > 793008506 */ 
+/* en > Barack Obama > 793008506 */
```
niedzielski: It would be awesome if you could add similar example articles like this for any of the rules you added or modified so it's easy to test where they apply later.

src/transform/ThemeTransform.js
nambatee: Sure, that sounds great, thanks @niedzielski!

montehurd:
Some of the issues I noticed, per my comment above, on theme demo articles, as a result of the changes not being scoped to respective chart/graph templates. There may be some dupes as I just took screenshots as I scrolled through the demo articles (there are MCS and Mobileview versions of each article). Also note that some percentage of these issues may have the same or related causes:

screen shot 2017-11-03 at 1 23 34 pm
screen shot 2017-11-03 at 1 23 41 pm
screen shot 2017-11-03 at 1 20 20 pm
screen shot 2017-11-03 at 1 20 35 pm
screen shot 2017-11-03 at 1 20 59 pm
screen shot 2017-11-03 at 1 23 52 pm
screen shot 2017-11-03 at 1 21 25 pm
screen shot 2017-11-03 at 1 21 36 pm
screen shot 2017-11-03 at 1 21 42 pm
screen shot 2017-11-03 at 1 21 47 pm
screen shot 2017-11-03 at 1 21 57 pm
screen shot 2017-11-03 at 1 22 03 pm
screen shot 2017-11-03 at 1 17 18 pm
screen shot 2017-11-03 at 1 17 25 pm
screen shot 2017-11-03 at 1 17 36 pm
screen shot 2017-11-03 at 1 17 41 pm
screen shot 2017-11-03 at 1 18 05 pm
screen shot 2017-11-03 at 1 18 12 pm
screen shot 2017-11-03 at 1 18 28 pm
screen shot 2017-11-03 at 1 18 34 pm
screen shot 2017-11-03 at 1 22 09 pm
screen shot 2017-11-03 at 1 22 16 pm
screen shot 2017-11-03 at 1 18 54 pm
screen shot 2017-11-03 at 1 19 29 pm
screen shot 2017-11-03 at 1 23 17 pm
screen shot 2017-11-03 at 1 19 47 pm
screen shot 2017-11-03 at 1 23 20 pm
screen shot 2017-11-03 at 1 13 30 pm
screen shot 2017-11-03 at 1 14 14 pm
screen shot 2017-11-03 at 1 14 39 pm
screen shot 2017-11-03 at 1 14 48 pm
screen shot 2017-11-03 at 1 15 03 pm
screen shot 2017-11-03 at 1 15 10 pm
screen shot 2017-11-03 at 1 15 19 pm
screen shot 2017-11-03 at 1 15 50 pm
screen shot 2017-11-03 at 1 16 06 pm
screen shot 2017-11-03 at 1 16 11 pm
screen shot 2017-11-03 at 1 16 16 pm
screen shot 2017-11-03 at 1 16 22 pm
screen shot 2017-11-03 at 1 16 27 pm
screen shot 2017-11-03 at 1 16 59 pm
screen shot 2017-11-03 at 1 20 10 pm
screen shot 2017-11-03 at 1 23 26 pm

montehurd:
@nambatee After I get the (native) headless transform performance tests in place I'd like to circle back and add tests for all of the theming issues in the screenshots above so you and @joewalsh can freely hack with immediate theme breakage feedback.

PR #97, opened 2017-11-03
Remove recursion from table collapse transform

montehurd:
I'd definitely like to further simplify and cleanup this XF, but for now I kept this PR diff minimal.

According to Safari Web Inspector Timelines recording, with this PR, on my iPhone 6s the `collapseTables` transform on `enwiki > Barack Obama` drops from 200 - 220ms to 2 - 8ms :)

### DIFF
--- a/src/transform/CollapseTable.js
+++ b/src/transform/CollapseTable.js
@@ -15,21 +15,19 @@ const SECTION_TOGGLED_EVENT_TYPE = 'section-toggled'
 const getTableHeader = (element, pageTitle) => {
   const thArray = []
   const headers = Polyfill.querySelectorAll(element, 'th')
-  if (headers) {
-    for (let i = 0; i < headers.length; ++i) {
-      const header = headers[i]
-      const anchors = Polyfill.querySelectorAll(header, 'a')
-      if (anchors.length < 3) {
-        // Also ignore it if it's identical to the page title.
-        if ((header.textContent && header.textContent.length) > 0
-          && header.textContent !== pageTitle && header.innerHTML !== pageTitle) {
-          thArray.push(header.textContent)
-        }
+  for (let i = 0; i < headers.length; ++i) {
+    const header = headers[i]
+    const anchors = Polyfill.querySelectorAll(header, 'a')
+    if (anchors.length < 3) {
+      // Also ignore it if it's identical to the page title.
+      if ((header.textContent && header.textContent.length) > 0
+        && header.textContent !== pageTitle && header.innerHTML !== pageTitle) {
+        thArray.push(header.textContent)
       }
-      if (thArray.length === 2) {
-        // 'newCaption' only ever uses the first 2 items.
-        break
-      }
+    }
+    if (thArray.length === 2) {
+      // 'newCaption' only ever uses the first 2 items.
+      break
     }
   }
   return thArray
--- a/src/transform/CollapseTable.js
+++ b/src/transform/CollapseTable.js
@@ -16,7 +16,8 @@ const getTableHeader = (element, pageTitle) => {
   const thArray = []
   const headers = Polyfill.querySelectorAll(element, 'th')
   if (headers) {
-    headers.forEach(header => {
+    for (let i = 0; i < headers.length; ++i) {
+      const header = headers[i]
       const anchors = Polyfill.querySelectorAll(header, 'a')
       if (anchors.length < 3) {
         // Also ignore it if it's identical to the page title.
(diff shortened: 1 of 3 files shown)


src/transform/CollapseTable.js
```diff
@@ -13,48 +13,25 @@ const SECTION_TOGGLED_EVENT_TYPE = 'section-toggled'
...
+      const header = headers[i]
+      const anchors = Polyfill.querySelectorAll(header, 'a')
+      if (anchors.length < 3) {
         // Also ignore it if it's identical to the page title.
-        if ((el.innerText && el.innerText.length || el.textContent.length) > 0
-                && el.innerText !== pageTitle && el.textContent !== pageTitle
-                && el.innerHTML !== pageTitle) {
-          thArray.push(el.innerText || el.textContent)
+        if ((header.textContent && header.textContent.length) > 0
+          && header.textContent !== pageTitle && header.innerHTML !== pageTitle) {
```
niedzielski: I think we want to use textContent but beware the differences for this conditional and the call to thArray.push() below: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

src/transform/CollapseTable.js
```diff
@@ -13,48 +13,25 @@ const SECTION_TOGGLED_EVENT_TYPE = 'section-toggled'
...
         }
       }
-    }
-
-    // if it's a table within a table, don't worry about it
-    if (el.tagName === 'TABLE') {
-      continue
-    }
-
-    // todo: why do we need to recurse?
```
niedzielski: Nice!

src/transform/CollapseTable.js
```diff
@@ -13,48 +13,25 @@ const SECTION_TOGGLED_EVENT_TYPE = 'section-toggled'
  * @return {!Array<string>}
  */
 const getTableHeader = (element, pageTitle) => {
```
niedzielski: This method was ported pretty closely from the original Android implementation (I don't know who wrote it) and there's obviously FUNKINESS about it. You may want to ask one of the Android devs to verify functionality before merging or add a CollapseTransform demo if they don't have time.

src/transform/CollapseTable.js
```diff
@@ -13,48 +13,25 @@ const SECTION_TOGGLED_EVENT_TYPE = 'section-toggled'
...
-      // ok, we have a TH element!
-      // However, if it contains more than two links, then ignore it, because
-      // it will probably appear weird when rendered as plain text.
-      const aNodes = el.querySelectorAll('a')
-      // todo: these conditionals are very confusing. Rewrite by extracting a
-      //       method or simplify.
-      if (aNodes.length < 3) {
-        // todo: remove nonstandard Element.innerText usage
+  const thArray = []
+  const headers = Polyfill.querySelectorAll(element, 'th')
```
niedzielski: According to the JSDocs, this always returns a nonnull array. Please remove the conditional below or update the JSDocs if they're wrong. I believe you could also write this as `headers.forEach(header => { ... })` since the index isn't used elsewhere.

test/transform/CollapseTable.test.js
```diff
@@ -45,10 +45,10 @@ describe('CollapseTable', () => {
             assert.deepEqual(actual, [])
```
niedzielski: No idea what's expected for these unit tests. I just wrote them according to existing functionality and treated the code as a black box (...filled with garbage).

src/transform/CollapseTable.js
montehurd: Ya I pinged Dmitry and Michael and Dmitry didn't recall (been a long time hehe) and I think Michael was out. 

It took a while to dissect but I'm pretty confident I grok what it was doing and that my replacement should produce same effect. Been testing it on various pages and so far no divergence... fingers crossed ;)

src/transform/CollapseTable.js
montehurd: Ya I initially used `forEach` but it doesn't support 'break' for early exit.

I'll remove the conditional - good catch!

src/transform/CollapseTable.js
montehurd: Ya I read about the differences and was initially hesitant but `innerText` is just so horrible :) So far I haven't seen any ill effects...

src/transform/CollapseTable.js
montehurd: Conditional removed!

test/transform/CollapseTable.test.js
montehurd: hehe took me longer than I care to admit to figure out that the test were formerly passing only because Domino doesn't implement `innerText`

src/transform/CollapseTable.js
montehurd: Guessing the combination of `innerText` forcing layouts and the unnecessarily deep recursion was the cause of the 2 order of magnitude speedup once those were removed...

mdholloway:
LGTM.  Nice performance win!

PR #98, opened 2017-11-06
Fix: graphs and charts issues on dark & sepia mode

nambatee:
https://phabricator.wikimedia.org/T179127

Fixes out 8 of 12 graph & charts issues without any visible regression in the demo. Fixes are visible on https://en.wikipedia.org/wiki/Wikipedia:Graphs_and_charts and https://en.wikipedia.org/wiki/PepsiCo, dark & sepia mode.


src/transform/ThemeTransform.js
```diff
@@ -74,10 +74,19 @@ const classifyElements = element => {
...
-  Polyfill.querySelectorAll(element, 'div.color_swatch div, div[style*="position: absolute"]')
-    .forEach(div => {
-      div.classList.add(CONSTRAINT.DIV_DO_NOT_APPLY_BASELINE)
-    })
+  /* en > Wikipedia:Graphs_and_charts > 801754530 */
+  /* en > PepsiCo > 799739023 */
+  const selector = [
+    'div.color_swatch div',
+    'div[style*="position: absolute"]',
+    'table.infobox div[role*="img"] div[style*="background-color"]',
```
montehurd: This line will get false positives and should probably be removed. Removing it makes the 'Single statistic' graph not work, but this approach would only work in the case where a single statistic graph happened to be contained by an infobox in the exact way it is on the 'Graphs and charts' page. 

The other query selector items look good though! They're all scoped and more specific.

I confirmed that I don't see any regressions now with the other theme demo articles, either on Dark or Sepia.

src/transform/ThemeTransform.js
```diff
@@ -74,10 +74,19 @@ const classifyElements = element => {
   /* en > Away colours > 793128975 */
   /* en > Manchester United F.C. > 793244653 */
   /* en > Pantone > 792312384 */
-  Polyfill.querySelectorAll(element, 'div.color_swatch div, div[style*="position: absolute"]')
-    .forEach(div => {
-      div.classList.add(CONSTRAINT.DIV_DO_NOT_APPLY_BASELINE)
-    })
+  /* en > Wikipedia:Graphs_and_charts > 801754530 */
+  /* en > PepsiCo > 799739023 */
```
montehurd: `799739023` is a revision of the Pepsico's talk page (the number looked too low). I think `807406166` is the latest.

montehurd:
Would you mind updating the theme demo files to include the `Pepsico` and `Graphs and charts` articles by adding...

     {"lang": "en", "title": "Wikipedia:Graphs_and_charts", "revision": 801754530},
     {"lang": "en", "title": "PepsiCo", "revision": 807406166}

... to `DemoArticles/articles.json` then run `./update.js` in `DemoArticles/`?

`update.js` fetches mobileView and MCS for every article revision in `DemoArticles/articles.json`. 

Then just commit `articles.json` and the four new files update fetches (ignore the changes to the existing files - sometimes changes to existing data files happens when structural html or timestamp changes occur upstream):

	demo/DemoArticles/data/en.PepsiCo.807406166.MCS.json
	demo/DemoArticles/data/en.PepsiCo.807406166.MV.json
	demo/DemoArticles/data/en.Wikipedia:Graphs_and_charts.801754530.MCS.json
	demo/DemoArticles/data/en.Wikipedia:Graphs_and_charts.801754530.MV.json

Then when you run the theme demo you'll then see these in the left menu and we can proof against these files any time we make further changes.

Also keep in mind you can temporarily remove items from `articles.json` if you want to just see one 2 articles in the demo menu instead of all of them. The demo menu is built from `articles.json` too. No need to re-run `update` if you just temporarily want fewer items in the menu.

nambatee:
@montehurd thanks for the review! 👍 Would it be ok if I added the steps to add demo articles to readme?

montehurd:
@nambatee 
> Would it be ok if I added the steps to add demo articles to readme?

Oh that would rule! The steps may change a bit when that new demo I mentioned replaces some of the existing ones, but it's based on the same data file structure and we can further tweak the readme then if needed.

montehurd:
@nambatee oh could you add line to `changelog.md` for this?

montehurd:
@niedzielski This lgtm.

niedzielski:
@montehurd, @nambatee lgtm too :+1: I'm happy to merge and publish if you like but I figured you're much more in tune to the library clients right now. Let me know if you want me to push the buttons :]

PR #99, opened 2017-11-07
Update readme to Include adding new demo articles

nambatee:


readme.md
```diff
@@ -133,7 +133,15 @@ Babel transpiles ES6 syntax to ES5 except for...of, spread, array destructuring,
...
-`npm run -s build:watch` will monitor and automatically rebuild the JavaScript and CSS bundles when files change. When used with NPM linking, this allows the app to always pull the latest local changes. Demos are viewable by running `npm run -s dev` which also updates the browser instantly for changes. Note: these demos only run in Chrome and Chromium presently.
+`npm run -s build:watch` will monitor and automatically rebuild the JavaScript and CSS bundles when files change. When used with NPM linking, this allows the app to always pull the latest local changes.
+
+### Demo
+Demos are viewable by running `npm run -s dev` which also updates the browser instantly for changes. Note: these demos only run in Chrome and Chromium presently.
+
+#### Adding new demo article
+1. Add new article in `DemoArticles/articles.json`
+2. Run `./update.js` in `DemoArticles/`
+3. Commit `articles.json` and the two new article fetches (ignore the changes to the existing files - sometimes changes to existing data files happens when structural html or timestamp changes occur upstream)
```
niedzielski: @montehurd, is it correct to always ignore these other files?

readme.md
```diff
@@ -133,7 +133,15 @@ Babel transpiles ES6 syntax to ES5 except for...of, spread, array destructuring,
 https://babeljs.io/docs/usage/caveats/#polyfills
 
 ### Automatic incremental builds
-`npm run -s build:watch` will monitor and automatically rebuild the JavaScript and CSS bundles when files change. When used with NPM linking, this allows the app to always pull the latest local changes. Demos are viewable by running `npm run -s dev` which also updates the browser instantly for changes. Note: these demos only run in Chrome and Chromium presently.
+`npm run -s build:watch` will monitor and automatically rebuild the JavaScript and CSS bundles when files change. When used with NPM linking, this allows the app to always pull the latest local changes.
+
+### Demo
+Demos are viewable by running `npm run -s dev` which also updates the browser instantly for changes. Note: these demos only run in Chrome and Chromium presently.
```
niedzielski: I thought the demos worked on all modern browsers now? See https://github.com/wikimedia/wikimedia-page-library/pull/76#issuecomment-326356868.

montehurd: I think they work in Safari etc now, though perhaps not as speedily.

readme.md
montehurd: If you're just adding one file doing so keeps the diff to just what you're adding. I think its fine to update the other files as well as long as the commit message mentions that that's being done.

readme.md
```diff
@@ -133,7 +133,15 @@ Babel transpiles ES6 syntax to ES5 except for...of, spread, array destructuring,
 https://babeljs.io/docs/usage/caveats/#polyfills
 
 ### Automatic incremental builds
-`npm run -s build:watch` will monitor and automatically rebuild the JavaScript and CSS bundles when files change. When used with NPM linking, this allows the app to always pull the latest local changes. Demos are viewable by running `npm run -s dev` which also updates the browser instantly for changes. Note: these demos only run in Chrome and Chromium presently.
+`npm run -s build:watch` will monitor and automatically rebuild the JavaScript and CSS bundles when files change. When used with NPM linking, this allows the app to always pull the latest local changes.
+
+### Demo
+Demos are viewable by running `npm run -s dev` which also updates the browser instantly for changes.
+
+#### Adding new demo article
```
montehurd: > Adding new demo article

May want add a note that presently these are only used by the theme transform?

( The demo I'm working on which lets you apply any of the transforms to whole articles uses them too but it's still WIP... )

readme.md
niedzielski: Sorry @nambatee and @montehurd, party foul! I didn't realize until now that this was in the old readme and you were just reformatting a little D:! Thanks for this improvement! :+1:

readme.md
```diff
@@ -139,6 +139,7 @@ https://babeljs.io/docs/usage/caveats/#polyfills
 Demos are viewable by running `npm run -s dev` which also updates the browser instantly for changes.
 
 #### Adding new demo article
+Note: These are only used by the theme transform.
```
montehurd: These are only used by the theme transform *demo*.

PR #100, opened 2017-11-11
Ensure 'display: block' for widened image ancestors

mdholloway:
Fixes another image widening bug in the Android app. Non-mw-gallery
images, in HTML loaded from action=mobileview, may have a thumbinner div
ancestor with a 'display: flex' or 'display: -webkit-flex' attribute.
Either of these prevents the image from widening as intended. Setting
'display: block' in the widenImage transform in these instances should
fix it.

Phab: https://phabricator.wikimedia.org/T163738 (see Steps to reproduce >
Mobileview)

### DIFF
--- a/src/transform/WidenImage.js
+++ b/src/transform/WidenImage.js
@@ -25,6 +25,12 @@ const widenAncestors = el => {
     if (parentElement.style.float) {
       parentElement.style.float = 'none'
     }
+    // Override 'display: flex' when present so that the image can fully widen. This seems to be an
+    // issue only on Android and with mobileview HTML. See, e.g., the first image in:
+    // https://en.m.wikipedia.org/wiki/Picasso%27s_Blue_Period?oldid=806483472
+    if (parentElement.style.display) {
+      parentElement.style.display = 'block'
+    }
   }
 }
 


montehurd:
@mdholloway oh nice! i'll kick the tires...

src/transform/WidenImage.js
```diff
@@ -25,6 +25,9 @@ const widenAncestors = el => {
     if (parentElement.style.float) {
       parentElement.style.float = 'none'
     }
+    if (parentElement.style.display) {
+      parentElement.style.display = 'block'
```
niedzielski: What do you think about using `inline-block` here? I was thinking it might be safer than `block`. For example, if the image isn't set to 100% width for whatever reason, `inline-block` seems more likely to work than `block`.

src/transform/WidenImage.js
```diff
@@ -25,6 +25,9 @@ const widenAncestors = el => {
     if (parentElement.style.float) {
       parentElement.style.float = 'none'
     }
+    if (parentElement.style.display) {
```
niedzielski: I love that you've taken the trouble to identify a use case in the wild. Would you mind adding a reference here? I've been using a style like: https://en.m.wikipedia.org/wiki/Picasso's_Blue_Period?oldid=806483472.

montehurd:
@mdholloway @niedzielski I'm still not seeing the bug from the phab ticket [repro steps](https://phabricator.wikimedia.org/T163738#3222167) on iOS. Could you check to see if the bug goes away if you turn off the android mobile app extension CSS? Guessing some of the crufty css we want do away with there could be interfering?

src/transform/WidenImage.js
montehurd: Inline block may make tablet layouts get a bit funky...

montehurd:
@mdholloway Oh cool! Lemme know what you find! I'm still keen on retiring as much of the mobile app extension CSS as possible, so I'm totally willing to help or test anything on iOS if needed.

src/transform/WidenImage.js
mdholloway: Yeah, I'm trying to get the old Nexus 9 charged up to test this myself :)

montehurd:
@mdholloway [here](https://github.com/wikimedia/wikipedia-ios/blob/develop/Makefile#L52)'s what we're using on iOS:

> /load.php?debug=false&lang=en&only=styles&skin=vector&modules=skins.minerva.base.reset|skins.minerva.content.styles|ext.math.styles|mobile.app.pagestyles.ios

mdholloway:
I looked at your Makefile and styles.css and you are indeed loading styles from production with hacks mode CSS included, same as us, so no difference there.  

I cannot get this to widen correctly without overriding display: flex.  The behavior persists even when I copy/paste the load.php query from your comment.  After that I even tried loading without the pagelib stylesheet to try and further isolate the problem, with the same result.  

At this point I'm at a loss.  Maybe it's a Blink vs. Webkit thing?

mdholloway:
I tested the latest beta release on a Jolla smartphone (which can run Android apps through an Android compatibility layer) and the image widened correctly.  It seems likely it's some quirk of the Android WebView that's causing this.

src/transform/WidenImage.js
mdholloway: Updated the comment.

src/transform/WidenImage.js
mdholloway: I did some manual testing and this seems fine in practice.  In principle, though, wouldn't the effect of this be to allow other images to appear beside widened images?  Is that something we want?

mdholloway:
Ping @dbrant: Hey!  I'm down a rabbit hole on this, probably well out of proportion to its actual priority.  

The 'easy' way to fix this would be to stop loading the styles from [hacks.less](https://github.com/wikimedia/mediawiki-skins-MinervaNeue/blob/master/resources/skins.minerva.content.styles/hacks.less) that are causing it.  Hacks.less contains hacks specific to the production Wikimedia environment.  Delivery of the hacks.less styles is controlled by a configuration option, [MinervaApplyKnownTemplateHacks](https://github.com/wikimedia/mediawiki-skins-MinervaNeue/blob/master/skin.json#L495), that's false as a MediaWiki default but true in Wikimedia production.  To avoid them we'd have to go back to loading styles from Vagrant, though, which is much more fiddly and error-prone, so I'm not sure that's a good idea. 

Separately, while working on image widening issues I've more than once been duped by the behavior that our app does not widen images on 'metered' (cellular) connections.  Do we really need this behavior?  I think I'm the one who introduced it but especially in light of image lazy loading it may not be needed.  (Actually, I'm not sure it's even documented, so it might be confusing users as well.)  Among other things this means that images are never widened on an emulator (they're always on mocked cellular connections) which is extremely easy to forget when debugging.  Alternatively, I guess we could allow widening on all connections for dev builds or something, but less complexity is always better.

My recommendations in short:

1. If @montehurd and @niedzielski don't want to clutter the widenImages transform with this, then maybe let's just consider the remainder of [T163738](https://phabricator.wikimedia.org/T163738) a wontfix. 

2. Let's strongly consider widening images in all cases rather than suppressing widening on cellular connections.

dbrant:
A definite +1 for widening images regardless of connection quality, given that we now have lazy-loading, plus our option to disable images altogether.

montehurd:
@mdholloway So... it sounds like both of your recommendations are leaning towards not changing the widen images js?

mdholloway:
@montehurd If it's mergeable as-is I'd say let's go for it.  If not, I don't really have more time to spend on it.  It's a pretty minor bug and only affects non-MCS page loads.  I'd probably just close this PR in that case.

montehurd:
@dbrant 

> A definite +1 for widening images regardless of connection quality, given that we now have lazy-loading, plus our option to disable images altogether.

So sounds like maybe closing this for now (in anticipation of widening regardless) is the way to go?

dbrant:
sgtm

PR #103, opened 2017-11-27
Add adjustTables

joewalsh:
Allows for setting up tables for expand / collapse behavior without initially collapsing them.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,5 @@
+- New: adjustTables allows for setting up tables for expand / collapse behavior without initially collapsing them
+
 ### v4.7.4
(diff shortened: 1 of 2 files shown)


changelog.md
```diff
@@ -1,3 +1,5 @@
+- New: adjustTables allows for setting up tables for expand / collapse behavior without initially collapsing them
```
montehurd: Will want to add `v5.0.0` on the line above since this change isn't backward compatible.

joewalsh: It is backwards compatible, `collapseTables` is still available

montehurd: Oops my bad!

PR #102, opened 2017-11-17
Fix code highlighting theme-ing

montehurd:
https://phabricator.wikimedia.org/T180846

`enwiki > Lua (programming language) > Syntax`

# Before
screen shot 2017-11-17 at 12 48 11 pm
screen shot 2017-11-17 at 12 48 20 pm

# After
screen shot 2017-11-20 at 5 07 53 pm
screen shot 2017-11-20 at 5 08 01 pm

PR #106, opened 2017-11-29
Use dark code syntax hightlight for new black theme too.

montehurd:
Per https://github.com/wikimedia/wikimedia-page-library/pull/102#pullrequestreview-79678402

### DIFF
--- a/src/transform/ThemeTransform.css
+++ b/src/transform/ThemeTransform.css
@@ -184,6 +184,10 @@ styles with only the things which need tweaking.
 .pagelib_theme_sepia .content div.mw-highlight {
   background: #f0e6d6;
 }
+.pagelib_theme_black .content code.mw-highlight,
+.pagelib_theme_black .content div.mw-highlight {
+  background: #000;
+}
 
 /* 'mw-highlight' colors for dark background */
 /* From: https://raw.githubusercontent.com/richleland/pygments-css/master/native.css */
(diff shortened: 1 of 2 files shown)


src/transform/ThemeTransform.css
```diff
@@ -266,6 +270,87 @@ styles with only the things which need tweaking.
 .pagelib_theme_dark .content .mw-highlight span.vm { color: #40ffff } /* Name.Variable.Magic */
 .pagelib_theme_dark .content .mw-highlight span.il { color: #3677a9 } /* Literal.Number.Integer.Long */
 
+/* Same 'mw-highlight' colors for dark background are used for black too */
+/* From: https://raw.githubusercontent.com/richleland/pygments-css/master/native.css */
+.pagelib_theme_black .content .mw-highlight span.hll { background-color: #404040 }
+/* .pagelib_theme_black .content .mw-highlight  { background: #202020; color: #d0d0d0 } */
```
niedzielski: This does not seem to be commented out in the original or dark mode. Is it intentional? If so, why?

montehurd: Oh this was intentional. The background (for black) is a bit further up:
https://github.com/wikimedia/wikimedia-page-library/pull/106/files#diff-340d93103cd4c516e89429d258b93b35R189

But I left these in place and commented out in case we ever re-sync with the upstream `pygments` classes as reminders that this line isn't wanted. We're using all of the other lines otherwise.

montehurd: I'm open to not doing that ^ and just removing these lines and instead adding a comment instead.

niedzielski: Ah, makes sense! A comment would be welcome at an opportune time but it's nothing to fret about.

PR #109, opened 2017-12-01
Design tweak for collapsing tables in black theme.

dbrant:
The collapsible tables have a black background which is the same as the
article backround, which is making them hard to discern within the page.

This adds a thin gray border around collapsible tables, per design
discussions, to make the table more discoverable in black mode.

### DIFF
--- a/src/transform/CollapseTable.css
+++ b/src/transform/CollapseTable.css
@@ -88,10 +88,14 @@ div.pagelib_collapse_table_container {
   clear: both;
 }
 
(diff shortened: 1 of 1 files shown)


montehurd:
@dbrant Sorry I missed this! Testing now...

montehurd:
@dbrant Do you need this right away or can it wait until we cut next build? No worries either way!

dbrant:
nope, next build is fine!

PR #111, opened 2017-12-08
Fix image widening with lazy load on iOS

montehurd:
On iOS if we used the image widening transform, then used the lazy image loading transform, the image placeholders were not correctly widened. Turned out to be related to image widening not also widening the image anchor, though I'm not quite sure why this only manifests itself on iOS webkit.

[Screenshots and discussion](https://github.com/wikimedia/wikipedia-ios/pull/1971#discussion_r155354570).

https://phabricator.wikimedia.org/T175323

### DIFF
--- a/src/transform/WidenImage.js
+++ b/src/transform/WidenImage.js
@@ -39,11 +39,10 @@ const updateStyleValue = (style, key, value) => {
  * @return {void}
  */
 const updateExistingStyleValue = (style, key, value) => {
-  const existingValue = style[key]
-  if (!existingValue) {
-    return
+  const valueExists = Boolean(style[key])
+  if (valueExists) {
+    updateStyleValue(style, key, value)
   }
-  updateStyleValue(style, key, value)
 }
 
 /**
@@ -162,6 +161,7 @@ export default {
   maybeWidenImage,
   test: {
     shouldWidenImage,
+    updateExistingStyleValue,
     widenAncestors
   }
 }
--- a/src/transform/WidenImage.js
+++ b/src/transform/WidenImage.js
@@ -46,15 +46,15 @@ const updateExistingStyleValue = (style, key, value) => {
 }
 
 /**
- * Array of arrays of image widening CSS key/value pairs.
- * @type {Array}
+ * Image widening CSS key/value pairs.
+ * @type {Object}
  */
-const styleWideningKeysAndValues = [
-  ['width', '100%'],
-  ['height', 'auto'],
-  ['maxWidth', '100%'],
-  ['float', 'none']
-]
+const styleWideningKeysAndValues = {
+  width: '100%',
+  height: 'auto',
+  maxWidth: '100%',
+  float: 'none'
+}
 
 /**
  * Perform widening on an element. Certain style properties are updated, but only if existing values
@@ -63,8 +63,8 @@ const styleWideningKeysAndValues = [
  * @return {void}
  */
 const widenElementByUpdatingExistingStyles = element => {
-  styleWideningKeysAndValues
-    .forEach(keyAndValue => updateExistingStyleValue(element.style, ...keyAndValue))
+  Object.keys(styleWideningKeysAndValues)
+    .forEach(key => updateExistingStyleValue(element.style, key, styleWideningKeysAndValues[key]))
 }
 
 /**
@@ -73,8 +73,8 @@ const widenElementByUpdatingExistingStyles = element => {
  * @return {void}
  */
 const widenElementByUpdatingStyles = element => {
-  styleWideningKeysAndValues
-    .forEach(keyAndValue => updateStyleValue(element.style, ...keyAndValue))
+  Object.keys(styleWideningKeysAndValues)
+    .forEach(key => updateStyleValue(element.style, key, styleWideningKeysAndValues[key]))
 }
 
 /**
--- a/src/transform/WidenImage.js
+++ b/src/transform/WidenImage.js
@@ -22,7 +22,7 @@ const ancestorsToWiden = element => {
 
 /**
  * Sets style value.
- * @param {CSSStyleDeclaration} style
+ * @param {!CSSStyleDeclaration} style
  * @param {!string} key
  * @param {*} value
  * @return {void}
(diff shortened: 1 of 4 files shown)


src/transform/WidenImage.js
```diff
@@ -1,30 +1,99 @@
...
+/**
+ * Sets style value only if value for given key already exists.
+ * @param {HTMLElement.style} style
+ * @param {!string} key
+ * @param {*} value
+ * @return {void}
+ */
+const updateExistingStyleValue = (style, key, value) => {
+  const existingValue = style[key]
+  if (!existingValue) {
```
niedzielski: Feel free to ignore but if you negate this conditional, you can move `updateStyleValue()` into it and omit the `return`.

src/transform/WidenImage.js
```diff
@@ -1,30 +1,99 @@
...
+ * @param {*} value
+ * @return {void}
+ */
+const updateStyleValue = (style, key, value) => {
+  style[key] = value
+}
+
+/**
+ * Sets style value only if value for given key already exists.
+ * @param {HTMLElement.style} style
```
niedzielski: :+1: for JSDocs! However, I think this should be the type `CSSStyleDeclaration`?

src/transform/WidenImage.js
```diff
@@ -1,30 +1,99 @@
...
+ * @param {!string} key
+ * @param {*} value
```
niedzielski: I think this would be a nullable string?

src/transform/WidenImage.js
```diff
@@ -1,30 +1,99 @@
 import './WidenImage.css'
```
niedzielski: This change involves a significant refactor and functionality change. If you don't mind, please update our tests for the new functionality and add a link to one of the articles that exhibits the problem.

src/transform/WidenImage.js
```diff
@@ -1,30 +1,99 @@
...
-    }
-    if (parentElement.style.float) {
-      parentElement.style.float = 'none'
-    }
+const widenAncestors = element => {
+  ancestorsToWiden(element).forEach(widenElementByUpdatingExistingStyles)
+
+  // Without forcing widening on the parent anchor, lazy image loading placeholders
+  // aren't correctly widened on iOS for some reason.
+  const parentAnchor = elementUtilities.findClosestAncestor(element, 'a.image')
```
niedzielski: Should we just check if element.parentElement is an instanceof HTMLAnchorElement instead?

src/transform/WidenImage.js
```diff
@@ -1,30 +1,99 @@
...
+  }
+  updateStyleValue(style, key, value)
+}
+
+/**
+ * Array of arrays of image widening CSS key/value pairs.
+ * @type {Array}
+ */
+const styleWideningKeysAndValues = [
+  ['width', '100%'],
```
niedzielski: Instead of using an array of arrays with the spread operator, this code might be hardier if it's changed to an array of objects, each with key and value members, or even better, just an object since order doesn't matter:

```js
const overrides = {
  width: '100%',
  height: 'auto',
  maxWidth: '100%',
  float: 'none'
}
```

src/transform/WidenImage.js
```diff
@@ -1,30 +1,99 @@
...
+  ['height', 'auto'],
+  ['maxWidth', '100%'],
+  ['float', 'none']
+]
+
+/**
```
niedzielski: It's a matter of opinion but I think the previous implementation was easier to read because it was a dead simple 30 lines. In my estimation, widenElementByUpdatingStyles and widenElementByUpdatingStyles don't offer much value because they're each only used in one place and are their bodies are only two lines. Similarly, updateExistingStyleValue and updateStyleValue could be replaced with anonymous lambdas because they're super short and the code speaks for itself. I get that you're emphasizing that in one case the presence of a key is checked for and in another the property is unconditionally set but I think it would be clear enough without explicit function names. /My two cents :|

src/transform/WidenImage.js
montehurd: Was thinking it could also be a number...

src/transform/WidenImage.js
montehurd: Ah good point. I've been doing more "early" returns to keep nesting shallower, but in this case it's only one line.

src/transform/WidenImage.js
montehurd: Ah, ya that makes sense. No real need for spread fanciness...

src/transform/WidenImage.js
montehurd: `widenAncestors` wasn't going to play nice with the next commit, which needed to set same style values, but unconditionally, and not on ancestors, rather, on a specific element.

I hear ya though. I tried to keep the new bits ultra-dead simple, with declarative naming so it's easy to see "it's doing this to the ancestors and it's doing that to the anchor".

src/transform/WidenImage.js
montehurd: I had that initially at first... but I wasn't sure if there were ever cases where some sort of template delivered img wrapped with anchor with some filling in between... I'm fine with switching back if you think its ok...

src/transform/WidenImage.js
montehurd: Fixed!

src/transform/WidenImage.js
montehurd: Fixed!

src/transform/WidenImage.js
montehurd: Fixed!

src/transform/WidenImage.js
montehurd: Tests added!

montehurd:
@niedzielski 

Made a bunch updates - let me know if I missed anything! Thanks again for all the CR! :)

src/transform/WidenImage.js
```diff
@@ -1,30 +1,98 @@
...
+      break
+    }
+    widenThese.push(el)
+  }
+  return widenThese
+}
+
+/**
+ * Sets style value.
+ * @param {CSSStyleDeclaration} style
```
niedzielski: I think this should be nonnull.

montehurd: Good catch! Fixed!

montehurd:
@dbrant gave thumbs-up on irc. merge time!

PR #112, opened 2017-12-08
Add 3 items to COPY_ATTRIBUTES which are used by iOS.

montehurd:
[Background](https://github.com/wikimedia/wikipedia-ios/pull/1971#issuecomment-350352888).

https://phabricator.wikimedia.org/T175323

### DIFF
--- a/src/transform/LazyLoadTransform.js
+++ b/src/transform/LazyLoadTransform.js
@@ -15,7 +15,10 @@ const IMAGE_LOADED_CLASS = 'pagelib_lazy_load_image_loaded' // Download complete
 
 // Attributes copied from images to placeholders via data-* attributes for later restoration. The
 // image's classes and dimensions are also set on the placeholder.
-const COPY_ATTRIBUTES = ['class', 'style', 'src', 'srcset', 'width', 'height', 'alt']
+// The 3 data-* items are used by iOS.
+const COPY_ATTRIBUTES = ['class', 'style', 'src', 'srcset', 'width', 'height', 'alt',
+  'data-file-width', 'data-file-height', 'data-image-gallery'
+]
 
 // Small images, especially icons, are quickly downloaded and may appear in many places. Lazily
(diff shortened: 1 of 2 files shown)


src/transform/LazyLoadTransform.js
```diff
@@ -15,7 +15,10 @@ const IMAGE_LOADED_CLASS = 'pagelib_lazy_load_image_loaded' // Download complete
 
 // Attributes copied from images to placeholders via data-* attributes for later restoration. The
 // image's classes and dimensions are also set on the placeholder.
-const COPY_ATTRIBUTES = ['class', 'style', 'src', 'srcset', 'width', 'height', 'alt']
+// The 3 data-* items are used by iOS.
+const COPY_ATTRIBUTES = ['class', 'style', 'src', 'srcset', 'width', 'height', 'alt',
+  'data-file-width', 'data-file-height', 'data-image-gallery'
```
niedzielski: This seems like a pretty harmless change. It's entirely up to your discretion but you may want to update tests for these new attributes. We currently validate all others and don't have to worry about breaking things so much when a change is made.

Just an aside and not a request for revision: on line 77, these attributes will be copied to `data-data-*` which is a bit clumsy. On the other hand, they can be handled generically without any other changes which is great.

montehurd: Ya I had to account for the `data-data` prefix when implementing tap handling https://github.com/wikimedia/wikipedia-ios/pull/1971/commits/8d50837b8da954cb3dfeef276f1fcb56b89eb44e on the placeholders (ie if a user taps the placeholder span it was nice to have that still try to show the native gallery for that placeholder's image). Agreed it's a bit clumsy, but figured we could circle back later if we refactor in future.

montehurd: @niedzielski Tests added :)

montehurd:
@dbrant gave thumbs-up on irc. merge time!

PR #115, opened 2017-12-13
Black theme tweaks

montehurd:
https://phabricator.wikimedia.org/T182726

# Before
screen shot 2017-12-12 at 5 56 19 pm

# After
screen shot 2017-12-12 at 5 55 36 pm

# Before
screen shot 2017-12-12 at 5 56 38 pm

# After
screen shot 2017-12-12 at 5 55 18 pm

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+### v4.7.10
+- Fix: follow-up "black" theme tweaks
+
 ### v4.7.9
 - Fix: encoding of titles when updating Read More save buttons
 
--- a/src/transform/CollapseTable.css
+++ b/src/transform/CollapseTable.css
@@ -95,7 +95,7 @@ div.pagelib_collapse_table_container {
 
 .pagelib_theme_black .content div.pagelib_collapse_table_container {
   box-shadow: none !important;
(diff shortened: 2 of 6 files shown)


montehurd:
@dbrant So I used black mode last night... I think I agree these changes are too bright for oled after dark... will show design on my X.

reetssydney:
hi @montehurd - just wanted to post her quick chat that I had with @dbrant that actually on Android the Black mode design *always intended* to use mid-background on quick facts boxes, so this change was in fact correcting an oversight on design review on my part. The idea is that it's not so egregious to use non-#000 since this box is always shown collapsed, and we would like some visual delineation over the main article content. 

Also the Android version of the footer uses 'Paper' (#000) for the read more and "About this article" so again there is less mid-background color usage.

The full article example is shown on the ticket https://phabricator.wikimedia.org/T173319  (file F11090992)

montehurd:
@reetssydney 

Understood. Just saying based on actually using it in the dark for the first time on an OLED screen, mid-background doesn't seem dark enough. On dark mode the visual hierarchy is great, but on black it just may need to be flattened a bit :)

PR #117, opened 2017-12-19
Fix image maps

joewalsh:
https://phabricator.wikimedia.org/T183288

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,5 @@
+- Fix: image maps were missing the usemap attribute after the lazy image load transform
+
 ### v4.7.11
 - Fix: more follow-up "black" theme tweaks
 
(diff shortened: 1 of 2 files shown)


changelog.md
```diff
@@ -1,3 +1,5 @@
+- Fix: image maps were missing the usemap attribute after the lazy image load transform
```
niedzielski: We normally add the next version here, so ` ### v4.7.12`.

joewalsh:
It's ok if we want to wait for other changes before cutting a new release.  I manually updated the js on the iOS side for now to get 5.7.3 out the door.

montehurd:
@niedzielski @joewalsh ah cool you guys have it covered (my ride doesn't leave for a while so I was doing a little CR)

PR #118, opened 2017-12-21
Fix for certain pages not getting sections appended

joewalsh:
On iOS, links from the sign post page weren't loading at all. 

Repro steps:

- In Safari, search for Wikipedia Signpost
- Tap the link for the sign post page (https://en.wikipedia.org/wiki/Wikipedia:Wikipedia_Signpost/Single/2017-12-18)
- Tap through to any of the links on the page

It's entirely possible these fixes are misguided, I just looked in the JS console and added protections around where there were errors for undefined values. After fixing these issues, the links from the signpost page load as expected.

### DIFF
--- a/src/transform/FooterReadMore.js
+++ b/src/transform/FooterReadMore.js
@@ -289,6 +289,9 @@ const updateSaveButtonBookmarkIcon = (button, isSaved) => {
 */
 const updateSaveButtonForTitle = (title, text, isSaved, document) => {
   const saveButton = document.getElementById(`${SAVE_BUTTON_ID_PREFIX}${encodeURI(title)}`)
+  if (!saveButton) {
+    return
+  }
   saveButton.innerText = text
   saveButton.title = text
   updateSaveButtonBookmarkIcon(saveButton, isSaved)
--- a/src/transform/FooterReadMore.js
+++ b/src/transform/FooterReadMore.js
@@ -289,9 +289,6 @@ const updateSaveButtonBookmarkIcon = (button, isSaved) => {
 */
 const updateSaveButtonForTitle = (title, text, isSaved, document) => {
   const saveButton = document.getElementById(`${SAVE_BUTTON_ID_PREFIX}${encodeURI(title)}`)
-  if (!saveButton) {
-    return
-  }
   saveButton.innerText = text
   saveButton.title = text
   updateSaveButtonBookmarkIcon(saveButton, isSaved)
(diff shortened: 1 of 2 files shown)


src/transform/FooterReadMore.js
```diff
@@ -289,6 +289,9 @@ const updateSaveButtonBookmarkIcon = (button, isSaved) => {
 */
 const updateSaveButtonForTitle = (title, text, isSaved, document) => {
   const saveButton = document.getElementById(`${SAVE_BUTTON_ID_PREFIX}${encodeURI(title)}`)
+  if (!saveButton) {
```
niedzielski: @montehurd, should this state be possible? Can we make it impossible?

src/transform/WidenImage.js
```diff
@@ -12,7 +12,7 @@ const ancestorsToWiden = element => {
   while (el.parentNode) {
     el = el.parentNode
     // No need to walk above 'content_block'.
-    if (el.classList.contains('content_block')) {
+    if (el.classList && el.classList.contains('content_block')) {
```
niedzielski: Can we change the loop to use `el.parentElement` instead? `classList` should always be defined for Elements but not necessarily Nodes.

https://developer.mozilla.org/en-US/docs/Web/API/Element
https://developer.mozilla.org/en-US/docs/Web/API/Node

src/transform/WidenImage.js
```diff
@@ -28,6 +28,9 @@ const ancestorsToWiden = element => {
  * @return {void}
  */
 const updateStyleValue = (style, key, value) => {
+  if (!style || !key) {
```
niedzielski: The JSDocs indicate these states are impossible when called correctly. Maybe this is a symptom of us using `el.parentNode` instead of `el.parentElement` since only Elements have a style property?

src/transform/WidenImage.js
montehurd: @joewalsh It looks like using `el.parentElement` on lines 12 and 13 fixes the problem. I'm looking at the save button console error message as part of the lazy loading fix. Would you be ok with changing this PR to just change lines 12 and 13?

montehurd:
Per @niedzielski's suggestion (and with @joewalsh's ok) I updated this PR to use `parentElement` vs `parentNode`.

montehurd:
https://phabricator.wikimedia.org/T183473

PR #119, opened 2018-01-03
Fix occassional console error when updating read more save buttons.

montehurd:


montehurd:
@joewalsh We could have round-tripped from native to JS first to see if the title currently had an onscreen read more item, but it just seemed simpler to make it safe to call even if no item is currently present.

PR #121, opened 2018-01-10
New: CollectionUtilities methods for gathering page issues and disambiguation information

montehurd:
https://phabricator.wikimedia.org/T164309

### DIFF
--- a/src/transform/CollectionUtilities.js
+++ b/src/transform/CollectionUtilities.js
@@ -14,9 +14,9 @@ const collectPageIssues = (document, element) => {
     Polyfill.querySelectorAll(element, 'table.ambox:not(.ambox-multiple_issues):not(.ambox-notice)')
   // Get the tables into a fragment so we can remove some elements without triggering a layout
   const fragment = document.createDocumentFragment()
-  const appendTableToFragment =
+  const cloneTableIntoFragment =
     table => fragment.appendChild(table.cloneNode(true)) // eslint-disable-line require-jsdoc
-  tables.forEach(appendTableToFragment)
+  tables.forEach(cloneTableIntoFragment)
   // Remove some elements we don't want when "textContent" or "innerHTML" are used
   Polyfill.querySelectorAll(fragment, '.hide-when-compact, .collapsed').forEach(el => el.remove())
   return Polyfill.querySelectorAll(fragment, 'td[class*=mbox-text] > *[class*=mbox-text]')
--- a/src/transform/CollectionUtilities.js
+++ b/src/transform/CollectionUtilities.js
@@ -1,0 +1,76 @@
+import Polyfill from './Polyfill'
+
+/**
+ * Extracts array of page issues from element
+ * @param {!Document} document
+ * @param {?Element} element
+ * @return {!Array.<string>} Return empty array if nothing is extracted
+ */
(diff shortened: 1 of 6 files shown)


test/transform/CollectionUtilities.test.js
```diff
@@ -0,0 +1,102 @@
+import assert from 'assert'
```
niedzielski: Woot for tests!

src/transform/CollectionUtilities.js
```diff
@@ -0,0 +1,76 @@
...
+const collectPageIssues = (document, element) => {
+  if (!element) {
+    return []
+  }
+  const tables =
+    Polyfill.querySelectorAll(element, 'table.ambox:not(.ambox-multiple_issues):not(.ambox-notice)')
+  // Get the tables into a fragment so we can remove some elements without triggering a layout
+  const fragment = document.createDocumentFragment()
+  const appendTableToFragment =
+    table => fragment.appendChild(table.cloneNode(true)) // eslint-disable-line require-jsdoc
```
niedzielski: Are we wanting to show page issues in multiple places? Just curious why this isn't a transform (_remove_ the tables from the passed in document and operate on them directly instead of cloning them).

test/transform/CollectionUtilities.test.js
```diff
@@ -0,0 +1,102 @@
...
+describe('CollectionUtilities', () => {
+  beforeEach(() => {
+    document = fixtureIO.documentFromFixtureFile('CollectionUtilities.html')
+  })
+
+  describe('.collectPageIssuesText()', () => {
+    it('find text issues', () => {
+      const element = document.querySelector('div#content_block_0')
+      assert.deepEqual(CollectionUtilities.collectPageIssuesText(document, element), [
+        'This article includes a list of references, but its sources remain unclear because it has insufficient inline citations.  (January 2016)', // eslint-disable-line max-len
```
niedzielski: There are different pros and cons and test styles but I'll just mention that these assertions would be simpler if the fixture was simplified (more unit-y) and less real world. No action is needed here but I wanted to share my thoughts.

src/transform/CollectionUtilities.js
montehurd: These utilities for the gathering part. I think there's an upcoming meeting ( @coreyfloyd ) to further iron out presentation of things like article footer menu across iOS, Android & mobile web... (on iOS for example, `Page issues` only appear in the article footer menu, but `Disambiguation` appears both inline and in the menu, which I think differs a bit from Android and mobile web)

test/transform/CollectionUtilities.test.js
montehurd: Good point. I was a bit lazy here :) These were mainly an easy check for me that my changes weren't breaking anything. I'll have a go at adding some more atomic tests...

src/transform/CollectionUtilities.js
montehurd: Oh regarding cloning into the fragment, the frag is just a convenient "scratchpad" we can use to easily remove elements we don't want represented when doing `innerHTML` or `textContent` (without removing them from the actual live document).

montehurd: I renamed the cloning `const` to be more declarative.

PR #125, opened 2018-01-23
Breaking: remove 'fragment' parameter from redlinks transform.

montehurd:
No longer needed on iOS. (we still use a fragment, but they can be 'enriched' to relay 'createElement' calls to the document on which they were created so the fragment can work as the document parameter for this transform's kickoff method)

### DIFF
--- a/src/transform/RedLinks.js
+++ b/src/transform/RedLinks.js
@@ -12,16 +12,15 @@ const configureRedLinkTemplate = (span, anchor) => {
 }
 
 /**
- * Finds red links in a document or document fragment.
- * @param {!(Document|DocumentFragment)} content Document or fragment in which to seek red links.
+ * Finds red links in a document.
+ * @param {!Document} content Document in which to seek red links.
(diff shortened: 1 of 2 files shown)


src/transform/RedLinks.js
```diff
@@ -12,16 +12,15 @@ const configureRedLinkTemplate = (span, anchor) => {
 }
 
 /**
- * Finds red links in a document or document fragment.
- * @param {!(Document|DocumentFragment)} content Document or fragment in which to seek red links.
```
niedzielski: Please add a release note for this API-breaking change.

montehurd: Ok if I update the changelog separately with note for this and https://github.com/wikimedia/wikimedia-page-library/pull/122, https://github.com/wikimedia/wikimedia-page-library/pull/123 and 
https://github.com/wikimedia/wikimedia-page-library/pull/124 ?

montehurd: That way I can just cut one build having all 4 updates.

niedzielski: :+1: Sounds good.

PR #127, opened 2018-01-30
Chore: add bundlesize test

niedzielski:
This patch adds production build output gzipped file size checks to the
test script so that any future efforts, including changes to the
bundler, may only introduce known size differences. Note:

• Initial limits were taken from the present gzipped chunk sizes rounded
  up to the 10th of a kibibyte.

• Only production JavaScript and CSS files are checked.

• The package.json bundlesize configuration options should be moved to a
  separate file when possible.

  https://github.com/siddharthkp/bundlesize/pull/171

• A leading zero is necessary for floating point limits less than one.
  Without a leading zero, checks against these limits appear to always
  pass.

• bundlesize was not added to dev because it doesn't support any quiet
  options and is verbose. Tests are only executed at precommit.

• JEDEC notation is used. bundlesize internally uses Bytes utility which
  only supports base 2 units.

  https://github.com/visionmedia/bytes.js

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "wikimedia-page-library",
-  "version": "4.5.6",
+  "version": "4.8.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


montehurd:
Coolnesss! Does this test run and print results like the other tests?

niedzielski:
@montehurd, yes! You should see it before you commit (precommit) or if you just want to run `npm -s run test:size`.

PR #131, opened 2018-04-05
Fix: edit pencil appearance

sharvaniharan:
**What Changed?**
-Changed edit pencil size from 28px to 24 px.
-Edit pencil now has its own table-cell space which floats to the right
-Edit container now aligns itself to the bottom

Screenshots for the change with different title lengths: https://drive.google.com/open?id=1sWwht34hLAVRYvnXa85CENnIQxOto4y-

Related Phabricator ticket: https://phabricator.wikimedia.org/T183018

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+### v5.0.2
+- Fix: Changing the size of the edit pencil to make it look more compatible with other edit pencil views in the android app. Also ensure that the edit pencil has its own dedicated column space within the header element, to support display with multiline headings. 
+
 ### v5.0.1
 - Chore: replace Rollup.js bundler with Webpack; there are no anticipated client changes necessary but integrators should smoke test CSS and JavaScript functionality, especially on older devices, and app devs should verify their development workflows
 - Chore: rename NPM `dev` script to canonical `start`
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,7 @@
-### v5.0.2
-- Fix: Changing the size of the edit pencil to make it look more compatible with other edit pencil views in the android app. Also ensure that the edit pencil has its own dedicated column space within the header element, to support display with multiline headings. 
+### v6.0.0
+- Breaking: update section header presentation for multiline titles; integration notes:
+  - Clients should now call EditTransform.newEditSectionHeader() instead of newEditSectionButton() and remove any custom code; please see JSDocs and demo
+- Update: change the size of the edit pencil to make it look more compatible with other edit pencil views in the android app
 
 ### v5.0.1
 - Chore: replace Rollup.js bundler with Webpack; there are no anticipated client changes necessary but integrators should smoke test CSS and JavaScript functionality, especially on older devices, and app devs should verify their development workflows
--- a/src/transform/EditTransform.css
+++ b/src/transform/EditTransform.css
@@ -4,8 +4,9 @@
 }
 
 .pagelib_edit_section_link_container {
-  margin: 2px 8px;
-  float: right;
+  margin: 2px 16px;
+  display: table-cell;
+  vertical-align: bottom;
   opacity: .54;
 }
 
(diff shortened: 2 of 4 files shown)


src/transform/EditTransform.css
```diff
@@ -4,8 +4,9 @@
 }
 
 .pagelib_edit_section_link_container {
-  margin: 2px 8px;
-  float: right;
+  margin: 2px 16px;
+  display: table-cell;
```
niedzielski: I would prefer to avoid usage of `display: table-cell` if possible because the rest of the content is not a tabular layout. I see that you've updated the release notes to indicate the problem this solves but I didn't quite follow! Would you mind updating the EditTransform demo with the use case? That would really help with testing and future development!

Here's what I see in the demo before:

screenshot from 2018-04-06 08-40-07

And after, which doesn't quite look right:

screenshot from 2018-04-06 08-43-40

changelog.md
```diff
@@ -1,3 +1,6 @@
+### v5.0.2
+- Fix: Changing the size of the edit pencil to make it look more compatible with other edit pencil views in the android app. Also ensure that the edit pencil has its own dedicated column space within the header element, to support display with multiline headings.
```
niedzielski: :+1:

niedzielski:
@sharvaniharan, I've made some amendments in a follow-up patch- feel free to revert. The patch consolidates some of the Android logic for creating a section header (I think everything but

niedzielski:
Sorry! Pressed the wrong button.

montehurd:
I'll try to do an iOS smoke test soon (still wrangling last reading list bits). Sorry for the delay!

sharvaniharan:
@niedzielski Works like a charm on Android! but the edit pencil has to be aligned to the last line of the title, and I added an attribute to set it... The h2,h3s look a pixel or so off on the single line titles after that addition, although it set right the alignment to bottom. Looking into it..

sharvaniharan:
Will put in the tests..

sharvaniharan:
Thanks @niedzielski for helping me out with the patch! Appreciate it!

sharvaniharan:
@montehurd Please merge if you feel everything's working on iOS.

montehurd:
@niedzielskis @sharvaniharan 

I just tried a trial iOS integration and it unfortunately doesn't seem to play nice with the way iOS uses the edit pencils.

On iOS the lead section pencil is like so:
screen shot 2018-04-20 at 12 34 33 pm

Vs non-lead section edit pencil:
screen shot 2018-04-20 at 12 34 41 pm

Any thoughts?

montehurd:
Specifically, the `section header` changes seem to have made it a bit less flexible for the iOS use case for seen in the first image.

montehurd:
On my trial integration I'm also not seeing the edit pencils move to the left on RTL articles. Unsure if this is an artifact of my integration (perhaps I'm using it incorrectly?) or if there's a `flex` issue on Safari?

sharvaniharan:
On Android, we were setting the direction on the app side, I removes that to get this working.

montehurd:
@sharvaniharan oh interesting! could you give me a link to the change?

sharvaniharan:
https://gerrit.wikimedia.org/r/#/c/425606/

montehurd:
Hmm don't have similar code on iOS - we just set the `HTML` tag dir and it cascaded down as far as I can tell.

Here's what I'm seeing:

screen shot 2018-04-20 at 1 03 54 pm

sharvaniharan:
Oh I had the exact same problem. Did you try not setting the dir html tag?

montehurd:
@sharvaniharan you were right about `row-reverse` causing it to double-reverse :) looking into fixes...

montehurd:
@sharvaniharan @niedzielski I pushed a commit with a fix for the RTL issue I was seeing. I almost have the iOS integration finished - realized I could still use the old `newEditSectionButton` func for the iOS lead section (from the 1st image I posted above).

sharvaniharan:
@montehurd Now, the only way it works is when i remove the reversal here and set the tag on android. I will remove it here... not sure why ...

sharvaniharan:
@montehurd as of now everything works... I will investigate on the tag separately... so if it is all working at your end.. feel free to merge.

PR #132, opened 2018-04-23
Hide iOS edit pencils for sections H3 and above.

montehurd:
6.0 changed the edit pencil containment a bit. Caused the CSS which hides iOS edit pencils on H3 and above to not behave itself.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+### v6.0.1
+- Fix: hiding edit pencils on iOS for sections H3 to H6
+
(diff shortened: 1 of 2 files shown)


niedzielski:
Published as NPM package v6.0.1. Please update your integration patch to reference this version (and don't forget to unlink before building the bundle).

https://www.npmjs.com/package/wikimedia-page-library
https://github.com/wikimedia/wikimedia-page-library/releases/tag/v6.0.1

PR #139, opened 2018-05-07
Fix `Read more` display titles (used only by iOS presently)

montehurd:


package.json
```diff
@@ -85,7 +85,7 @@
     },
     {
       "path": "build/wikimedia-page-library-transform.js",
-      "maxSize": "12.7KB"
+      "maxSize": "12.73KB"
```
niedzielski: I normally just bump up to the next tenth of a kilobyte but this is fine too :+1:

PR #140, opened 2018-05-09
New: reference collection

montehurd:
Used to make it easier to show native adjacent reference panel carousel:

dino mov

# Usage details:
Used on iOS when a reference is tapped to show a native panel which lets user drag side to side to see any info for references which were adjacent to the tapped reference. ie: if a lone reference is tapped ie `[4]` we show a panel with its reference text, but if that reference has adjacent references ie `[3][4][5]` the panel we show can be dragged side to side to see the references text for `[3]` and `[5]`.

To use:
Pass `isCitation` the href string of the tapped reference anchor, if it returns true (on iOS we use it to determine if the [type of the clicked item is `reference`](https://github.com/wikimedia/wikipedia-ios/pull/2295/files#diff-88236b5102c629ba152585c1a3fd753bR32)) call `collectNearbyReferences` with the document and the tapped anchor element to extract relevant references info ([iOS example](https://github.com/wikimedia/wikipedia-ios/pull/2295/files#diff-88236b5102c629ba152585c1a3fd753bR134)) including a `referencesGroup` array and the index of the selected element `selectedIndex` from that group.

The each item in `referencesGroup` will contain:
 - `id` - id of the reference
 - `rect` - [ClientRect](https://www.javascripture.com/ClientRect) of the reference (so you can do a native overlay dimming everthing but the tapped reference link ie `[4]`)
 - `text` - the text of the tapped reference link ie `[4]`
 - `html` - the html/text of the reference from the actual reference at the bottom of the page (in the animation above this is everything shown below the horizontal line in the panels)

# iOS native interface details:
- On iOS we use it as outlined above to extract adjacent reference info, and we send that info [across](https://github.com/wikimedia/wikipedia-ios/pull/2295/files#diff-88236b5102c629ba152585c1a3fd753bR136) the iOS version of the JS/Native bridge. 
- We then present a native interface showing the adjacent reference group, with one panel per reference (the bottom of the panel has a web view for showing the reference html, but it just looks like part of the panel, as you can see in the animation). 
- The native code also applies an overlay with a cutout around the tapped ref. 
- It also scrolls the web view up if the native panel it presents would cover the tapped reference, as you can see in the animation when I tap reference `[11]`.

# Notes: 
- I also moved some utilities to `NodeUtilities.js` - apologies that this wasn't as separate commit.
- I had to fairly heavily refactor the iOS implementation to modernize it a bit and make it more testable.
- On iOS on tablet we only show the single tapped reference (no side-to-side sliding panels or dark overlay):
screen shot 2018-05-09 at 4 33 00 pm


sharvaniharan:
Thankyou @montehurd :) yay! i'm looking forward to use this!

montehurd:
@sharvaniharan You're welcome! I just edited the description with a few more details of how we use this on iOS. Let me know if you have any questions or issues :)

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+/**
+ * Get the reference text container.
+ * @param  {!Document} document
+ * @param  {!Node} sourceNode
+ * @return {?HTMLElement}
+ */
+const getRefTextContainer = (document, sourceNode) => {
+  const refTextContainerID = getFirstChildAnchor(sourceNode).getAttribute('href').slice(1)
+  const refTextContainer = document.getElementById(refTextContainerID)
+    || document.getElementById(decodeURIComponent(refTextContainerID))
```
niedzielski: Do we really need to check both decoded and undecoded?

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+  if (Polyfill.matchesSelector(sourceNode, REFERENCE_SELECTOR)) {
+    return sourceNode
+  }
+  return ElementUtilities.findClosestAncestor(sourceNode, REFERENCE_SELECTOR)
+}
+
+/**
+ * Reference item model.
+ */
+class ReferenceItem {
```
niedzielski: What do you think about using a JSDoc for typing instead of a class? Then creating a ReferenceItem is just matching the shape which requires explicit labeling of properties. For example, before:

```
class ReferenceItem {
  constructor(id, rect, text, html) {
    this.id = id
    this.rect = rect
    this.text = text
    this.html = html
  }
}
const reference = new ReferenceItem('a', {width, height}, 'foo', '<b>bar</b>')
```

With labeled properties:
```
/**
 * @typedef ReferenceItem
 * @prop {string} id
 * @prop {DOMRect} rect
 * @prop {?string} text
 * @prop {?string} html
 */
const reference = {id: 'a', rect: {width, height}, text: 'foo', html: '<b>bar</b>'}
```

Same thing for NearbyReferences.

test/transform/ReferenceCollection.test.js
```diff
@@ -0,0 +1,266 @@
...
+      assert.ok(isCitation('#') === false)
+    })
+  })
+
+  describe('.adjacentNonWhitespaceNode()', () => {
+    const adjacentNonWhitespaceNode = ReferenceCollection.test.adjacentNonWhitespaceNode
+    beforeEach(() => {
+      document.documentElement.innerHTML = `
+        <stuff>
+          \t<b id=one>one</b>
```
niedzielski: Sorry but why an escaped tab here?

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+ * @param  {!Collector} nodeCollector
+ * @return {void}
+ */
+const collectAdjacentReferenceNodes = (node, siblingGetter, nodeCollector) => {
+  let currentNode = node
+  while (true) {
+    currentNode = adjacentNonWhitespaceNode(currentNode, siblingGetter)
+    if (hasCitationLink(currentNode)) {
+      nodeCollector(currentNode)
+      continue
```
niedzielski: If you invert this condition, you can replace the continue here with the break below.

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+ * Collect adjacent reference nodes.
+ * @param  {!Node} node
+ * @param  {!SiblingGetter} siblingGetter
+ * @param  {!Collector} nodeCollector
+ * @return {void}
+ */
+const collectAdjacentReferenceNodes = (node, siblingGetter, nodeCollector) => {
+  let currentNode = node
+  while (true) {
+    currentNode = adjacentNonWhitespaceNode(currentNode, siblingGetter)
```
niedzielski: You may wish to explicate in the JSDocs that the starting node is never collected.

sharvaniharan:
@montehurd works on Android now! @niedzielski will let you merge once the changes you have requested are in!

niedzielski:
Re-review pending follow-ups from @montehurd.

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+ * Determines if node is a text node containing only whitespace.
+ * @param  {!Node}  node
+ * @return {!boolean}
+ */
+const isWhitespaceTextNode = node =>
+  !(!node || node.nodeType !== Node.TEXT_NODE || !node.textContent.match(/^\s+$/))
+
+/**
+ * Checks if element has a child anchor with a citation link.
+ * @param  {!HTMLElement}  element
```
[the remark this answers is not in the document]
montehurd: Fixed

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+ * @return {!boolean}
+ */
+const hasCitationLink = element => {
+  try {
+    return isCitation(getFirstChildAnchor(element).getAttribute('href'))
+  } catch (e) {
```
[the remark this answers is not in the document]
montehurd: Hmm I don't recall the exact reason for the `try/catch` - been a while. I played with your alternative for a bit (adding call to `isCitation` where needed) but didn't get it to work (tests failed). Would be ok to leave this as-is for now?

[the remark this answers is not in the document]
montehurd: Oh nevermind I think I fixed it...

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+/**
+ * Gets first child anchor.
+ * @param  {!HTMLElement} element
+ * @return {?HTMLAnchorElement}
+ */
+const getFirstChildAnchor = element => element.querySelector('A')
+
+/**
+ * Determines if node is a text node containing only whitespace.
+ * @param  {!Node}  node
```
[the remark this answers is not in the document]
montehurd: Fixed

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+import Polyfill from './Polyfill'
+
+const REFERENCE_SELECTOR = '.reference'
+
+/**
+ * Is Citation.
+ * @param  {!string}  href
+ * @return {!boolean}
+ */
+const isCitation = href => href.indexOf('#cite_note') > -1
```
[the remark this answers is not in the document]
montehurd: I think I prefer to keep the method narrowly scoped vs passing it the entire anchor element. Open to refactoring this later if you feel strongly about it though :)

I added and used a constant per your suggestion. For using the anchor rather than the `href` string

src/transform/ReferenceCollection.js
montehurd: Perhaps no longer, but I do recall some of these had funky encoding at one time...

test/transform/ReferenceCollection.test.js
montehurd: The test is for `adjacentNonWhitespaceNode()`, so iirc I separated the elements with newlines and leading space "whitespace", so I thought it would be good to throw tabs in there too.

src/transform/ReferenceCollection.js
montehurd: Doh! Good catch! Fixed

src/transform/ReferenceCollection.js
montehurd: Ah good point. Added

montehurd:
@niedzielski I think I replied to and/or pushed commits to all your comments. Should be good for a re-review when you have time. Thanks again for the great feedback!

sharvaniharan:
@niedzielski @montehurd can we merge this?

src/transform/ReferenceCollection.js
```diff
@@ -0,0 +1,253 @@
...
+ * @return {!boolean}
+ */
+const isCitation = href => href.indexOf(CITE_HASH_PREFIX) > -1
+
+/**
+ * Gets first child anchor.
+ * @param {!HTMLElement} element
+ * @return {?HTMLAnchorElement}
+ */
+const getFirstChildAnchor = element => element.querySelector('A')
```
niedzielski: Heya!

I wanted to share my perspective on replacing comments with functions as you've done here.

In general, I think it's great practice to use good naming on our symbols and, of course, abstraction for the win. However, I probably wouldn't write a comment for querying the first anchor of a DOM subtree since it's so commonplace as to be un-noteworthy and the abstraction here does nothing. Function calls may now read more like English but they also obscure the underlying simplicity of what's actually happening here with a new custom DSL and greatly inflate the verbosity of the code which is more of a decrease to readability than an increase.

Consider what happens also when the second child anchor is needed in some new patch. Do we make a new function that sits alongside getFirstChildAnchor() called getSecondChildAnchor()? That seems bad enough but when the third anchor is needed, suddenly we might wonder if getNChildAnchor() should be added which takes an argument, n. Now we have three functions that all boil down to querySelector / querySelectorAll and we still need to pass an argument.

I think we both feel very comfortable working with querySelector / querySelectorAll so my opinion is that we should favor their direct usage unless we're doing something extra that we want to capture in a function.

test/transform/ReferenceCollection.test.js
```diff
@@ -0,0 +1,266 @@
...
+        <sup id="cite_ref-b" class="reference"><a id='a2' href="#BLA">[5]</a></sup>
+      `)
+    })
+    it('correctly affirms child anchor has citation link', () => {
+      const element = document.querySelector('#cite_ref-a')
+      assert.equal(hasCitationLink(element), true)
+    })
+    it('correctly affirms child anchor does not have citation link', () => {
+      const element = document.querySelector('#cite_ref-b')
+      assert.equal(hasCitationLink(element), false)
```
niedzielski: I replaced all these calls to assert.equal with assert.strictEqual since for the same reason we favor `===` to `==`. :+1:

src/transform/ReferenceCollection.js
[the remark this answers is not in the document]
niedzielski: Perfect! I'm glad we have a test case that fails! Now we can easily make the improvement right now and be confident in our code instead of punting it down the line and be confused later.

I looked into this myself. This try / catch block was obscuring the real underlying issue which was that adjacentNonWhitespaceNode() can return null and getFirstChildAnchor() expects nonnull.

src/transform/ReferenceCollection.js
niedzielski: And that when there is only one node, the passed in collector is never called.

montehurd: Doesn't it go without saying that if there no adjacent reference nodes, none will be collected?

montehurd:
@niedzielski Heya! Just a head's-up I reverted 22bfca3 as it didn't work. I'll take a peek at the cause and undo the revert if I can suss it out :)

src/transform/ReferenceCollection.js
montehurd: Ya this was overly abstracted. I removed it. Good catch :)

test/transform/ReferenceCollection.test.js
montehurd: Coolio! Thanks!

src/transform/ReferenceCollection.js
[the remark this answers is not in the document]
montehurd: Heya just a heads-up I reverted your change - see my comment here https://github.com/wikimedia/wikimedia-page-library/pull/140#issuecomment-396567201 for details.

niedzielski:
> Edit: So it looks like it's not working because you're trying to call querySelector on a text node. btw sorry if some of these bits were a little hard to follow - I've tried to make it more grok-able than the original but there's certainly room for further improvement :)
Great! Thanks for diagnosing. I've added a test case for the scenario you've described and as I understand the desired behavior, and made the fix. Please be sure to check it out!

test/transform/ReferenceCollection.test.js
```diff
@@ -0,0 +1,266 @@
+import assert from 'assert'
+import domino from 'domino'
+import pagelib from '../../build/wikimedia-page-library-transform'
+
+let document
+
+const ReferenceCollection = pagelib.ReferenceCollection
+
+const referenceGroupHTML = `
+  <sup id="cite_ref-a" class="reference"><a id='a1' href="#cite_note-a">[4]</a></sup>
```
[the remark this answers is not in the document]
montehurd: Fixed!

src/transform/ReferenceCollection.js
montehurd: > Usage of the class keyword exposes us to misusage of extends

Using classes to model data is pretty common. How is classes being extendable practically risky here?

montehurd:
@niedzielski Your 5ae2c77 commit worked! Thanks!

If you have a sec there are a couple questions above:
https://github.com/wikimedia/wikimedia-page-library/pull/140#discussion_r198659298
https://github.com/wikimedia/wikimedia-page-library/pull/140#discussion_r194708228

montehurd:
@niedzielski btw thanks again for all the *great* comments and CR!!!

sharvaniharan:
@niedzielski @montehurd merging this! thanks for all the review work you guys did!! 💯

montehurd:
I mentioned this PR on...
https://phabricator.wikimedia.org/T165432

PR #141, opened 2018-05-18
Fix: brick, pie, bar, and stacked bar chart theming

niedzielski:
https://phabricator.wikimedia.org/T179128


niedzielski:
@montehurd can you take a peek at these PRs? This one's already stale but doesn't have any feedback.

montehurd:
@niedzielski will do! i'm jumping back on these pagelib PRs today :)

montehurd:
This is great! Just noticed a couple small issues when proofing in dark mode with the theme demo:

- Looks like the single statistic bar color isn't being excluded from override:
screen shot 2018-06-01 at 4 27 39 pm
screen shot 2018-06-01 at 4 31 13 pm

- Also wondering if we could/should exclude the color of the text overlaying these bars:
screen shot 2018-06-01 at 4 27 47 pm
screen shot 2018-06-01 at 4 31 20 pm

niedzielski:
@montehurd, ready for review!

> single statistic bar color
This is a Template:Composition_bar which requires the templateeditor permission to edit and I do not have it yet. While I want to see this composition bar issue fixed, I included as many other bonus fixes as I could as part of this series of patches and I believe it covers the original bug reported.

> Also wondering if we could/should exclude the color of the text overlaying these bars:
The CSS for this isn't pretty but "fixed" :]

PR #142, opened 2018-05-18
Fix: chessboard theming

niedzielski:
https://phabricator.wikimedia.org/T183333


montehurd:
Approved but we'll probably want to modify this PR's addition to the changelog since `v6.0.8` is already out.

PR #143, opened 2018-05-19
Fix: infobox detection on frwiki

niedzielski:
Bug: T188096


src/transform/CollapseTable.js
```diff
@@ -84,8 +84,9 @@ const isNodeBreakElement = node => node.nodeType === ELEMENT_NODE && node.tagNam
  * @param  {!Node} node
  * @return {void}
  */
-const replaceNodeWithBreakingSpaceTextNode = (document, node) =>
+const replaceNodeWithBreakingSpaceTextNode = (document, node) => {
```
montehurd: Ah, guessing these braces are so nothing gets returned...

src/transform/FooterReadMore.js
```diff
@@ -218,7 +217,7 @@ const stringFromQueryParameters = parameters => Object.keys(parameters)
  * @param {!string} title
  * @param {!number} count Number of `Read more` items to fetch for this title
  * @param {?string} baseURL
- * @return {!sring}
```
montehurd: Haha good catch!

PR #145, opened 2018-06-08
Update package-lock

mdholloway:
Needed to update growl (a dependency of mocha), which has a security
vulnerability at 1.9.2.

https://github.com/mochajs/mocha/issues/2973
https://nodesecurity.io/advisories/146

PR #148, opened 2018-06-15
Fix: generate source maps for debug builds

niedzielski:
Source maps were not generated for debug builds. There seem to be two
issues:

  - The setting in the Webpack configuration file does not appear to be
    used. It must be specified on the command line.

  - cheap-module-eval-source-map doesn't seem to work so use the same
    thing as prod, source-map.

### DIFF
--- a/webpack.config.babel.js
+++ b/webpack.config.babel.js
@@ -63,7 +63,7 @@ const config = {
 
   stats: STATS,
 
-  devtool: PRODUCTION ? 'source-map' : 'cheap-module-eval-source-map',
+  devtool: 'source-map',
 
   devServer: PRODUCTION ? undefined : {
     clientLogLevel: 'warning',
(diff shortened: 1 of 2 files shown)


webpack.config.babel.js
```diff
@@ -63,7 +63,7 @@ const config = {
 
   stats: STATS,
 
-  devtool: PRODUCTION ? 'source-map' : 'cheap-module-eval-source-map',
+  devtool: 'source-map',
```
berndsi: Ran into this, too. Was about to suggest the same.

niedzielski:
Published as NPM package [v6.0.9](https://www.npmjs.com/package/wikimedia-page-library/v/6.0.9). Please update your integrations to reference this version (and don't forget to unlink before building the bundle). /cc @berndsi, @mdholloway, @montehurd, @dbrant, changes included:

https://github.com/wikimedia/wikimedia-page-library/blob/master/changelog.md#v609
https://github.com/wikimedia/wikimedia-page-library/pull/145
https://github.com/wikimedia/wikimedia-page-library/pull/147
https://github.com/wikimedia/wikimedia-page-library/pull/148

PR #150, opened 2018-06-25
Make the edit pencil in section titles optional.

dbrant:
In certain cases, we don't actually want specific sections to be editable (whether due to template transclusion, Parsoid oddities, etc), so we'd like to be able to optionally show or hide the edit pencil when building the section header element.

### DIFF
--- a/test/transform/EditTransform.test.js
+++ b/test/transform/EditTransform.test.js
@@ -62,4 +62,22 @@ describe('EditTransform', () => {
       assert.equal(element.childNodes.item(0).nodeName, 'H4')
     })
   })
+
+  describe('.newEditSectionHeader(0, 2, false)', () => {
+    let document
+    let element
+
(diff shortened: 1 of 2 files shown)


test/transform/EditTransform.test.js
```diff
@@ -62,4 +62,22 @@ describe('EditTransform', () => {
       assert.equal(element.childNodes.item(0).nodeName, 'H4')
     })
   })
+
```
niedzielski: Thanks for the test!

niedzielski:
Published as NPM package [v6.0.10](https://www.npmjs.com/package/wikimedia-page-library/v/6.0.10). Please update your integrations to reference this version (and don't forget to unlink before building the bundle). /cc @dbrant, @montehurd, @mdholloway, @berndsi.

PR #151, opened 2018-07-12
LazyLoadTransformer now can handle existing placeholders

berndsi:
Adds a new method LazyLoadTransformer.collectExistingPlaceholders which
should be used instead of `convertImagesToPlaceholders` if the
placeholders have been created server-side.

The new method `convertImagesToPlaceholders` looks for elements with the placeholder class.
No need to call `convertImagesToPlaceholders`.

Bug: https://phabricator.wikimedia.org/T198668

### DIFF
--- a/src/transform/LazyLoadTransform.ts
+++ b/src/transform/LazyLoadTransform.ts
@@ -165,6 +165,7 @@ const loadPlaceholder = (document: Document, placeholder: HTMLSpanElement): HTML
 }
 
 export default {
+  PLACEHOLDER_CLASS,
   queryLazyLoadableImages,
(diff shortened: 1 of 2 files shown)


berndsi:
Sorry, the commit message in the first commit is wrong. Let me know if you want me to resubmit a new PR instead.

niedzielski:
Published as NPM package v6.1.1. Please update your integration to reference this version (and don't forget to unlink).

berndsi:
Updated the Gerrit patch to use the published version of page lib: https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/445073/5..6

PR #153, opened 2018-07-27
Add demo for CollapseTable transform

berndsi:
Also running ThemeTransform.classifyElements(), just in case.

### DIFF
--- a/demo/CollapseTable.html
+++ b/demo/CollapseTable.html
@@ -1,0 +1,650 @@
+<!doctype html>
+<html lang=en>
+
+  <head>
+    <meta charset=utf-8>
+    <title>CollapseTable demo</title>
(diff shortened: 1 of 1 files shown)


berndsi:
@niedzielski I just realized that the ThemeTransform demo has a checkbox for turning on CollapseTable. So, I guess this PR is not useful. Instead I've added the article to the Them demo (PR #154).
Unfortunately the RTL issue is present there as well.

niedzielski:
@berndsi, up to you. The theming demo is enormous and takes a while to load so debugging issues is slower. I'm happy to have some overlap with the demos, especially if this is more unit focused.

berndsi:
@niedzielski Yeah, the Theme demo is super slow. I wonder how hard it would be to lazy load the article content.

berndsi:
Maybe the "all" case could be solved by adding a button to load all content? But by default only load the first one, so a user can set all the control parameters (theme, collapseTables, direction, ...) before loading all articles.

niedzielski:
That would be a nice improvement.

montehurd:
> Maybe the "all" case could be solved by adding a button to load all content? But by default only load the first one, so a user can set all the control parameters (theme, collapseTables, direction, ...) before loading all articles.

This is more or less what the "everything demo" I'm chipping away at does. Sorry it's taking so long!

berndsi:
@montehurd working on a better demo.

PR #158, opened 2018-08-24
Avoid infobox too narrow issue

berndsi:
This change undoes some of
https://gerrit.wikimedia.org/r/c/mediawiki/skins/MinervaNeue/+/428868
but only for tables we collapse, and therefore don't float.

QA: Turn on Collapse tables and compare expanded infoboxes in:
http://localhost:3000/ThemeTransform.html#en.Chris%20Cornell.793305307.MCS.json
http://localhost:3000/ThemeTransform.html#en.Cat.792971769.MCS.json

Bug: https://phabricator.wikimedia.org/T201381

### DIFF
--- a/src/transform/CollapseTable.css
+++ b/src/transform/CollapseTable.css
@@ -172,3 +172,11 @@ div.pagelib_collapse_table_container {
 .pagelib_collapse_table_icon {
   box-sizing: border-box;
 }
+
+/* Avoid infobox too narrow issue, see https://phabricator.wikimedia.org/T201381,
+   undoing some of https://gerrit.wikimedia.org/r/c/mediawiki/skins/MinervaNeue/+/428868
+   but only for tables we collapse. */
+.pagelib_collapse_table_container table {
+    display: block;
+    width: 100% !important
+}


montehurd:
Oh cool I'll try to kick the tires on this later today! :)

montehurd:
Heya! I kicked the tires on iOS but I can't seem to find any tables which look differently. Maybe I goofed something... I check `Cat` and `Chris Cornell` but they look the same... could you maybe take an Android screenshot of one of the affected tables? There may be some weird iOS specific thing we're doing... will keep looking...

montehurd:
Is the `200px inline` mentioned on the ticket maybe not happening with mobileview?

berndsi:
@montehurd The width of the screen needs to be > 720px or you won't see the problem in the released version of the iOS app. I did see it on a friends iPhone when in landscape mode.

montehurd:
I tried landscape on the pages I checked... hmm...

montehurd:
I was using iPhone 6 simulator... lemme try 6 Plus...

montehurd:
@berndsi That was it! I'm seeing it on the 6 Plus! Thanks!!

berndsi:
Here's a screenshot on master with the Theme demo on the Cat article using mobileview.
Note that "Collapse tables" needs to be checked and the browser window needs to be wide. Then you open the infobox and scroll down a bit:

screen shot 2018-08-27 at 5 07 45 pm

berndsi:
Didn't see your previous comments that you were able to repro until now. Yes, this was on an iPhone 6 Plus in landscape mode.

src/transform/CollapseTable.css
```diff
@@ -171,4 +171,12 @@ div.pagelib_collapse_table_container {
...
   box-sizing: border-box;
-}
\ No newline at end of file
+}
+
+/* Avoid infobox too narrow issue, see https://phabricator.wikimedia.org/T201381,
+   undoing some of https://gerrit.wikimedia.org/r/c/mediawiki/skins/MinervaNeue/+/428868
+   but only for tables we collapse. */
+.pagelib_collapse_table_container table {
+    display: block;
```
berndsi: We could also use `display: table` here. Not sure which one is better.

PR #160, opened 2018-09-24
Add Header elements

berndsi:
Need to make a decision on I18N strings. Here we only have one string that needs to be translated. If we do the Footer as well there would be a bunch more. Should we setup I18N for the page library?


berndsi:
Replaced by #163

PR #164, opened 2018-10-31
First pass at 'everything demo' of transforms on actual articles

montehurd:
https://phabricator.wikimedia.org/T172166

This first pass implements the following:

- [x] apply transforms to actual articles (lazy image loading not yet implemented)
- [x] use same article data files as theme demo
- [x] allow multiple articles to be viewed at once
- [x] allow quick-jump between articles being viewed
- [x] automatically re-apply transform selections to newly selected articles

It still needs to conform to our linting and documentation standards, which I'll chip away at.

@berndsi and I also discussed updates which will need to be made to support representation of `PCS` article data. Things we discussed:
- [x] a way to make `articleRef` support non-section loading (PCS delivers [entire article](https://en.wikipedia.org/api/rest_v1/page/mobile-html/Dog), not sections - move index html and section html wrapping into `articleRef`?) and related changes to the demo html file
- [ ] <strike>a way to exclude particular transforms from particular sources (for example, PCS already does certain transforms, MCS already relocates 1st paragraphs, etc...)</strike> (Will look at later)
- [ ] <strike>what we could show above each article preview iframe to indicate what transforms were actually applied to that article (PCS for example wouldn't list any transforms it applies upstream) to make proofing expectations easier to confirm</strike> (Will look at later)

Other considerations:
- should we eventually remove some of the old demo code? or only, say, all demos except the lazy image loading demos?


berndsi:
@montehurd This looks great. Just one question: Is the order the transforms show up in the control in the same order as they should be invoked? If not I think that would be useful. Other than this question this looks great.

PR #163, opened 2018-10-29
Add Header Transform

berndsi:
This PR adds a new transform to add a page header, including title, possible description, possible pronunciation button, edit button (for lead section and possibly description as well), and a decorative divider.

One I18N string is passed in assuming I18N is handled on the PCS side.

Also added a new demo for this with the various options.

https://phabricator.wikimedia.org/T201383

PR #166, opened 2018-11-10
New edit tranform method for article header

montehurd:
Article header turned out to be simpler to implement as a special case of a section header, which the edit transform already supported.

### DIFF
--- a/src/transform/EditTransform.css
+++ b/src/transform/EditTransform.css
@@ -105,3 +105,57 @@ a.pagelib_edit_section_link {
 .pagelib_platform_ios h6.pagelib_edit_section_title + span.pagelib_edit_section_link_container {
     display: none;
 }
+
+/* Wikidata description styling */
+#pagelib_edit_section_title_description {
+    font-size: 14px;
+    line-height: 22px;
+    color: #777D84;
+    margin-top: 3px;
+    margin-bottom: 5px;
+    padding-bottom: 0px;
+}
+
+#pagelib_edit_section_add_title_description {
+    font-size: 14px;
+    font-style: italic;
+    display: inline;
+}
+
+#pagelib_edit_section_add_title_description:before {
+    content: "";
+    background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACAAAAAMCAYAAAADFL+5AAAAAXNSR0IArs4c6QAAADFJREFUOBFjNE47859hAAHTANoNtnrUAQMdA6P2D3wIMFK7HDg7y4SRFG+NZsMBDwEAJj8Gc0QaQK8AAAAASUVORK5CYII=');
+    background-repeat: no-repeat;
+    width: 1.8em;
+    height: 0.75em;
+    background-size: 1.0em 0.375em;
+    display: inline-block;
+    background-position: left center;
+}
+
+[dir="rtl"] #pagelib_edit_section_add_title_description:before {
+    transform: scaleX(-1);
+}
+
+/* Horizontal divider between title/description and lead section text */
+#pagelib_edit_section_divider {
+    border: 1px solid #ccc;
+    margin-top: 12px;
+    margin-bottom: 16px;
+    width: 60px;
+    margin-left: 0px;
+}
+
+@media (-webkit-min-device-pixel-ratio: 2) {
+    #pagelib_edit_section_divider {
+        border-width: 0.5px;
+    }
+}
+
+[dir="ltr"] #pagelib_edit_section_divider {
+    margin-left: 0px;
+}
+
+[dir="rtl"] #pagelib_edit_section_divider {
+    margin-right: 0px;
+}
--- a/src/transform/EditTransform.js
+++ b/src/transform/EditTransform.js
@@ -8,6 +8,12 @@ const CLASS = {
   PROTECTION: { UNPROTECTED: '', PROTECTED: 'page-protected', FORBIDDEN: 'no-editing' }
 }
 
+const IDS = {
+  TITLE_DESCRIPTION: 'pagelib_edit_section_title_description',
+  ADD_TITLE_DESCRIPTION: 'pagelib_edit_section_add_title_description',
+  DIVIDER: 'pagelib_edit_section_divider'
+}
+
 const DATA_ATTRIBUTE = { SECTION_INDEX: 'data-id', ACTION: 'data-action' }
 const ACTION_EDIT_SECTION = 'edit_section'
 
@@ -47,9 +53,10 @@ const newEditSectionButton = (document, index) => {
  * @param {!number} level The *one-based* header or table of contents level.
  * @param {?string} titleHTML Title of this section header.
  * @param {?boolean} showEditPencil Whether to show the "edit" pencil (default is true).
+ * @param {?string} anchor Section anchor used for jumping between sections.
  * @return {!HTMLElement}
  */
-const newEditSectionHeader = (document, index, level, titleHTML, showEditPencil = true) => {
+const newEditSectionHeader = (document, index, level, titleHTML, showEditPencil = true, anchor) => {
   const element = document.createElement('div')
   element.className = CLASS.SECTION_HEADER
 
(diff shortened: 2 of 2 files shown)


montehurd:
I still need to grab the pronunciation bits from https://github.com/wikimedia/wikimedia-page-library/pull/163

src/transform/EditTransform.js
```diff
@@ -64,11 +71,63 @@ const newEditSectionHeader = (document, index, level, titleHTML, showEditPencil
...
+ * Lead section header is a special case as it needs to show article title and description too,
+ * and in addition to the lead edit pencil, the description can also be editable.
+ * As a client, you may wish to set the ID attribute.
+ * @param {!Document} document
+ * @param {?string} articleDisplayTitle Article display title.
+ * @param {?string} titleDescription Article title description.
+ * @param {?string} addTitleDescriptionString Localized string e.g. 'Add title description'.
+ * @param {?boolean} isTitleDescriptionEditable Whether title description is editable.
+ * @param {?boolean} showEditPencil Whether to show the "edit" pencil (default is true).
+ * @param {?string} anchor Section anchor used for jumping between sections.
```
berndsi: Wouldn't this be '#'?

src/transform/EditTransform.js
```diff
@@ -64,11 +71,63 @@ const newEditSectionHeader = (document, index, level, titleHTML, showEditPencil
...
+ * @return {!HTMLElement}
+ */
+const newEditLeadSectionHeader = (document, articleDisplayTitle, titleDescription,
+  addTitleDescriptionString, isTitleDescriptionEditable, showEditPencil = true, anchor) => {
```
berndsi: When would this be false?

src/transform/EditTransform.js
```diff
@@ -64,11 +71,63 @@ const newEditSectionHeader = (document, index, level, titleHTML, showEditPencil
...
+    // TODO: consider renaming this 'id' to 'anchor' for clarity - would need to update native
+    // code as well - used when TOC sections made to jump to sections.
+    element.id = anchor
+  }
+
   return element
 }
 
+/**
+ * Lead section header is a special case as it needs to show article title and description too,
```
berndsi: This code should also work for any kinds of pages, but just article pages. Would you mind changing here wherever it says article to page?

src/transform/EditTransform.js
montehurd: On iOS right now the lead section edit pencil doesn't appear directly to the right of the article title. It's further down. I left the parameter in for now so iOS can immediately start using this transform even before we switch over to PCS.

berndsi: Ah, I see. Thanks!

src/transform/EditTransform.js
montehurd: Will do!

montehurd: Fixed!

src/transform/EditTransform.js
montehurd: checking...

montehurd: iirc - `id` from, say, mobile view, is just a number, which isn't specific enough to use because there could be cases where the section content could also contain an element with a numeric `id`. that's why mobile view also provides `anchor`, which should be more safe(r) to use for section container id (which is used for jumping to section).

montehurd: i suppose the parameter to be renamed... but i guess i thought it made more sense to leave it named the same as the api param...

montehurd: Removing this - on iOS we can add this after as-needed.

montehurd: Fixed! Good catch!

src/transform/EditTransform.js
```diff
@@ -8,8 +8,17 @@ const CLASS = {
...
 
+const IDS = {
+  TITLE_DESCRIPTION: 'pagelib_edit_section_title_description',
+  ADD_TITLE_DESCRIPTION: 'pagelib_edit_section_add_title_description',
+  DIVIDER: 'pagelib_edit_section_divider',
+  PRONUNCIATION: 'pagelib_edit_section_title_pronunciation'
+}
+
 const DATA_ATTRIBUTE = { SECTION_INDEX: 'data-id', ACTION: 'data-action' }
 const ACTION_EDIT_SECTION = 'edit_section'
```
cooltey: Is it possible to have a const for the lead edit section?
 e.g. `const ACTION_EDIT_LEAD_SECTION = 'edit_lead_section'`

The Wikipedia android will popup an overflow menu when clicking on the edit anchor in the lead section. Otherwise will need to override the attribute by scanning the whole HTML.

src/transform/EditTransform.css
```diff
@@ -105,3 +124,57 @@ a.pagelib_edit_section_link {
...
+    transform: scaleX(-1);
+}
+
+/* Horizontal divider between title/description and lead section text */
+#pagelib_edit_section_divider {
+    border: 1px solid #ccc;
+    margin-top: 12px;
+    margin-bottom: 16px;
+    width: 60px;
+    margin-left: 0px;
```
cooltey: should it be removed?

src/transform/EditTransform.css
```diff
@@ -105,3 +124,57 @@ a.pagelib_edit_section_link {
...
+}
+
+@media (-webkit-min-device-pixel-ratio: 2) {
+    #pagelib_edit_section_divider {
+        border-width: 0.5px;
+    }
+}
+
+[dir="ltr"] #pagelib_edit_section_divider {
```
cooltey: Should it be 
`
#pagelib_edit_section_divider[dir="ltr"] {
  xxx...
}
` 
since it must be applied if the #pagelib_edit_section_divider has the attritube called `dir`?

src/transform/EditTransform.css
```diff
@@ -105,3 +124,57 @@ a.pagelib_edit_section_link {
...
+    margin-left: 0px;
+}
+
+[dir="rtl"] #pagelib_edit_section_divider {
```
cooltey: Same

src/transform/EditTransform.css
cooltey: It should be removed since it will be applied in the following style.

src/transform/EditTransform.js
```diff
@@ -63,12 +72,86 @@ const newEditSectionHeader = (document, index, level, titleHTML, showEditPencil
...
+  }
+  if (isTitleDescriptionEditable) {
+    const a = document.createElement('a')
+    a.href = '#'
+    a.setAttribute(DATA_ATTRIBUTE.ACTION, ACTION_ADD_TITLE_DESCRIPTION)
+    const p = document.createElement('p')
+    p.id = IDS.ADD_TITLE_DESCRIPTION
+    p.innerHTML = addTitleDescriptionString
+    a.appendChild(p)
+    return a
```
cooltey: Is it possible to let the `p` to be the parent and make `p.appendChild(a)` and then return `p`? If so, we can set the `dir` attritube by selecting the `p` in any condition.

Also, if changes the logic above, you might need to remove the `display: inline` from `#pagelib_edit_section_add_title_description` to make the `dir` works.

src/transform/EditTransform.js
montehurd: When you detect the click event has 'data-action' of 'edit_section', you can check if 'data-id' is zero to see if the click is for the lead section. Or am I misunderstanding what you're trying to do?

cooltey: Thanks @montehurd!

The reason is that we have another action listener for the edit icon in the lead section. https://github.com/wikimedia/apps-android-wikipedia/blob/master/www/js/editaction.js#L9

It should be ok if stay as it is since we can also update the attribute manually in our `section.js`
https://gerrit.wikimedia.org/r/#/c/apps/android/wikipedia/+/476787/3/www/js/sections.js

montehurd: Ah ok sounds good! I'm looking at your other comments now but I may have to chip away at it over next couple days - also having to do some native work too :)

montehurd: Ya on [line 5](https://github.com/wikimedia/apps-android-wikipedia/blob/master/www/js/editaction.js#L5) it looks like you can just check if 'data-id' is 0 and send the bridge message from [line 10](https://github.com/wikimedia/apps-android-wikipedia/blob/master/www/js/editaction.js#L10).

src/transform/EditTransform.css
montehurd: fixed!

src/transform/EditTransform.css
montehurd: I think was some legacy stuff from before we switch to setting `dir` on the html tag. I removed it and confirmed expected alignment still works ok on iOS (for example, on `EN` the divider is left aligned and on `HE` it's right aligned).

montehurd: Checking on the other cases now...

montehurd: The other places I was using `dir` were actually needed - I [scoped them](https://github.com/wikimedia/wikimedia-page-library/pull/166/commits/1a6428beba7207f67ade52efe0c9490248aaa0c3) to the `html` tag (so these automatically flip horizontally in accordance with the `html` tag's direction) to avoid future confusion.

src/transform/EditTransform.css
montehurd: Fixed - see note [here](https://github.com/wikimedia/wikimedia-page-library/pull/166#discussion_r238895952).

src/transform/EditTransform.js
montehurd: Oh can you double-check that you still need to set `dir` - if I recall correctly Android sets `dir` on the `html` tag so you probably don't need to set it on nested elements such as this one...

montehurd: (I think we only set `dir` on the `html` tag on iOS these days...)

cooltey: Thanks @montehurd !
We have been set the `dir` tag to the `div` tag that has the class name `pagelib_edit_section_header`. 

But looks like set `dir` in the parent `div` (the `id='content'`) is a good idea and much simpler!

src/transform/EditTransform.css
cooltey: Solved!

cooltey:
All looks good to me!

montehurd:
@cooltey woohoo! thanks!

PR #174, opened 2019-02-14
Add .node-version

joewalsh:
For use with [nodenv](https://github.com/nodenv/nodenv)


niedzielski:
We're using [NVM](https://github.com/creationix/nvm) for the Readers Web repos. Do you know how nodenv compares? In the metric of Git stars, NVM is significantly more popular but I don't know that nodenv was even considered.

.node-version
```diff
@@ -0,0 +1 @@
+7.5.0
```
niedzielski: Should we drop the engine version from package.json to avoid synchronicity issues? Also, please consider updating the development set up notes in the readme!

joewalsh:
I compared [rbenv](https://github.com/rbenv/rbenv) and [rvm](https://github.com/rvm/rvm) several years ago and preferred [rbenv's approach](https://github.com/rbenv/rbenv/wiki/Why-rbenv%3F). From there I picked nodenv just by association with rbenv, so I can't give a good comparison to nvm. Nodenv does have a similar [self-promotional page](https://github.com/nodenv/nodenv/wiki/Why-nodenv%3F) about how they compare.

I don't think we should necessarily force everyone to switch over, but it would be nice to support both. It looks like `nvm` can use a `.nvmrc` file and we could symlink to that from `.node-version`

.node-version
joewalsh: Yeah it would be ideal to only have to update this in one place. Does anything else depend on the engine version?

niedzielski: I think you get a warning if you try to install with an older version of Node but that's it.

https://docs.npmjs.com/files/package.json#engines

joewalsh:
Closing for now - will re-investigate after https://github.com/wikimedia/wikimedia-page-library/pull/175

PR #176, opened 2019-02-25
Switch previous theme bug fix - 'unset' to 'initial'.

montehurd:
Android had a version compatibility issue with 'unset':
https://phabricator.wikimedia.org/T214728


montehurd:
The hot fix here https://gerrit.wikimedia.org/r/#/c/mediawiki/services/mobileapps/+/492751/ should fix the KitKat bug. This page lib PR just makes the page lib consistent with the change made by the hot fix.

montehurd:
...so should be no rush to cut and use updated page lib build using this PR.

PR #177, opened 2019-02-27
Fix absolutely positioned image widening bug

montehurd:
https://phabricator.wikimedia.org/T215667

### DIFF
--- a/test/fixtures/WidenImage.html
+++ b/test/fixtures/WidenImage.html
@@ -24,6 +24,11 @@
         </div>
       </div>
     </div>
+    <div style='position:absolute'>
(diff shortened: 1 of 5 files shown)


test/fixtures/WidenImage.html
```diff
@@ -24,6 +24,11 @@
         </div>
       </div>
     </div>
+    <div style='position:absolute'>
```
joewalsh: If this is written as `position: absolute`, which is valid css, this test fails

montehurd: @joewalsh Fixed!

PR #178, opened 2019-03-13
Pre-emptively override upstream CSS changes

joewalsh:
There are changes coming to the upstream CSS that will break the display of reference lists in the apps. It might also break other ordered lists, but haven't found another repro case. This change preemptively overrides that style to ensure it remains unchanged in the apps for now. Context:

https://phabricator.wikimedia.org/T150377#5019102

Screenshots with CSS manually updated to the upstream proposed change:

Screenshot_20190313-122011

Simulator Screen Shot - iPhone X - 2019-03-12 at 16 14 00

There are also changes to padding around captions pending, I've added an override for those as well based on this thread: 
https://phabricator.wikimedia.org/T168861#5019022

### DIFF
--- a/src/transform/CollapseTable.css
+++ b/src/transform/CollapseTable.css
@@ -82,6 +82,7 @@
 table.infobox caption {
   text-align: center;
   font-weight: bold;
+  padding: 16px 16px 8px 16px;
 }
 
 div.pagelib_collapse_table_container {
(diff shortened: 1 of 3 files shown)


sharvaniharan:
Although the references look ok, we just made an update to fix a bug on tablets. We stopped setting customized horizontal padding, since it was happening upstream. But now, it seems to be breaking with this change. We might just have to put it back. Will discuss this Monday with @dbrant and @cooltey  and update here:

reference


padding

sharvaniharan:
dark

sharvaniharan:
Above bug is unrelated to this change. So good to go on Android! 👍

PR #181, opened 2019-03-29
Allow client to set margin/padding on body element

berndsi:
This commit introduces a new transform, called BodySpacingTransform.
It allows client to set margins and padding on the body element of the
given HTML Document.

Bug: https://phabricator.wikimedia.org/T201382

### DIFF
--- a/src/transform/BodySpacingTransform.ts
+++ b/src/transform/BodySpacingTransform.ts
@@ -7,25 +7,25 @@ interface Spacing {
 
 /**
  * Sets the margins on an element via inline styles.
- * @param {!HTMLElement} element the element that needs the margins adjusted.
+ * @param {!HTMLBodyElement} bodyElement the element that needs the margins adjusted.
  *   For the apps this is usually the body element.
  * @param {Spacing} values { top, right, bottom, left }
  *   Use value strings with units, e.g. '16px'. Undefined values are ignored.
  * @param callback callback function
  * @return {void}
  */
-const setMargins = (element: HTMLElement, values: Spacing, callback?: () => void): void => {
+const setMargins = (bodyElement: HTMLBodyElement, values: Spacing, callback?: () => void): void => {
   if (values.top !== undefined) {
-    element.style.marginTop = values.top
+    bodyElement.style.marginTop = values.top
   }
   if (values.right !== undefined) {
-    element.style.marginRight = values.right
+    bodyElement.style.marginRight = values.right
   }
   if (values.bottom !== undefined) {
-    element.style.marginBottom = values.bottom
+    bodyElement.style.marginBottom = values.bottom
   }
   if (values.left !== undefined) {
-    element.style.marginLeft = values.left
+    bodyElement.style.marginLeft = values.left
   }
   if (callback) {
     callback()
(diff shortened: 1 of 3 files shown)


src/transform/BodySpacingTransform.ts
```diff
@@ -0,0 +1,65 @@
...
+    body.style.marginRight = values.right
+  }
+  if (values.bottom !== undefined) {
+    body.style.marginBottom = values.bottom
+  }
+  if (values.left !== undefined) {
+    body.style.marginLeft = values.left
+  }
+  if (callback) {
+    callback()
```
montehurd: Wondering if we can do without the callback here and in `setPadding` (since they're not passed anything and they both happen at very end of their respective methods)? i.e. could you just directly call anything which needs to happen after `setMargins` or `setPadding` by invoking that code after calls to `setMargins` / `setPadding`?

berndsi: I suspect this is true. I mainly added it because it was requested in the ticket. @joewalsh : thoughts?

joewalsh: The thought was to add a `requestAnimationFrame` before the callback to ensure the update was rendered before the callback is called - this way the apps could adjust the margins on launch and not remove the cover until after they've changed. Same with the theme and other visual methods with callbacks.

berndsi: @montehurd Any other concerns?

src/transform/BodySpacingTransform.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Sets the margins on the body element via inline styles.
+ * @param {!Document} document
+ * @param {Spacing} values { top, right, bottom, left }
+ *   Use value strings with units, e.g. '16px'. Undefined values are ignored.
+ * @param callback callback function
+ * @return {void}
+ */
+const setMargins = (document: Document, values: Spacing, callback?: () => void): void => {
+  const body: HTMLElement = document.body
```
montehurd: May consider just passing these the body element since document isn't used for anything else beyond getting the body...

test/transform/BodySpacingTransform.test.js
```diff
@@ -0,0 +1,50 @@
...
+      const document = domino.createDocument('<p></p>')
+      BodySpacingTransform.setMargins(document, {})
+      assert.strictEqual(document.body.outerHTML, '<body><p></p></body>')
+    })
+
+    it('just left + right', () => {
+      const document = domino.createDocument('<p></p>')
+      BodySpacingTransform.setMargins(document, { right: '8px', left: '16px' })
+      const style = 'margin-right: 8px; margin-left: 16px;'
+      assert.strictEqual(document.body.outerHTML,`<body style="${style}"><p></p></body>`)
```
montehurd: This could fail in future if domino changes the whitespace it uses to separate inline styles components, or the component order. May consider separate asserts for expected `document.body.style.marginRight` and `document.body.style.marginLeft` values.

montehurd: Same for other asserts below...

src/transform/BodySpacingTransform.ts
montehurd: @berndsi I added a couple small notes. Not must-haves but may be nice.

src/transform/BodySpacingTransform.ts
berndsi: We could make this transform more generic but then it adds more burden to the caller. Ok, maybe not so much. I was actually on the fence about it. If you feel like we should then I'd probably rename this transform to not have `body` in the name anymore. Let me know.

montehurd: Oh, I think it's fine to have 'body' in the name and the 'document' param renamed to 'body'.

test/transform/BodySpacingTransform.test.js
berndsi: Done

src/transform/BodySpacingTransform.ts
berndsi: Done

PR #182, opened 2019-04-11
Remove background-color rule

berndsi:
Undoes the fix for T214728. The negative side effect for this is that
it caused issues with mobile-html.

I don't believe this rule is still needed now that the PCS base CSS
removes most of the unneeded CSS rules from Minerva.

Bug: https://phabricator.wikimedia.org/T217837


montehurd:
Looks like mobileview on iOS is unhappy without the bg color rule:
Screen Shot 2019-04-18 at 1 35 19 PM

Could this be because we need to update our CSS source [url](https://github.com/wikimedia/wikipedia-ios/blob/develop/Makefile#L26)?
> `CSS_ORIGIN = $(CSS_PREFIX)/load.php?only=styles&target=mobile&skin=minerva&modules=skins.minerva.base.styles|skins.minerva.content.styles|ext.math.styles|ext.pygments|mobile.app`

joewalsh:
iOS issue fixed by https://github.com/wikimedia/wikipedia-ios/pull/3053

PR #187, opened 2019-04-17
Update to current long term stable node version (10.15.3)

joewalsh:
- Adds `.node-version` and `.nvmrc` files instead of the `engines` entry in `package.json`. `.nvmrc` is a symlink to `.node-version` so only one change is required to update.
- Ran npm audit fix to upgrade dependencies with security issues

1 high level vulnerability remains in the version of `tar` that is used by the devDependency `bundlesize`.


joewalsh:
@berndsi is [v6.11.1](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/package.json#L97) still the node version used for mobileapps services? Would it help to downgrade pagelib to 6.11.1 to match instead of upgrading to 10.15.3?

berndsi:
> @berndsi is [v6.11.1](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/package.json#L97) still the node version used for mobileapps services? 

Yes. mobileapps still uses Node 6 until we move to Kubernetes. But that should happen in this or next quarter.

> Would it help to downgrade pagelib to 6.11.1 to match instead of upgrading to 10.15.3?

It would add a small convenience of me not having to switch node versions so often but that's really minor. I think it's better to stay with Node 10 at this point, esp. now that we have plans to [move to Node 10 in mobileapps](https://phabricator.wikimedia.org/T218733) soon.

package.json
```diff
@@ -54,7 +51,7 @@
     "babel-preset-env": "1.7.0",
     "babel-register": "6.26.0",
     "browser-sync": "2.26.3",
-    "bundlesize": "0.17.0",
+    "bundlesize": "^0.17.1",
```
niedzielski: If this were pinned to a patch version, it would match the other dependencies.

joewalsh: Done

joewalsh:
>I think some of the tests have Node.js system dependencies but, given that this library is otherwise a concatenation of local JavaScript files, do you think the version of Node.js itself matters much? I think it is the dependency versions that matter and since the package-lock file is committed and supported by newer versions of Node.js, is it necessary to specify Node.js version? It doesn't hurt anything to have it but I thought I'd ask.

The package versions in the package-lock definitely matter more, but I see Node.js as another dependency of this repo. It's helpful to at least try to keep that variable constant across different dev machines even if it doesn't currently matter for the built output. I verified that the build output is identical between 7.5.0 and 10.15.3 when the same package-lock file is used, though I'm not sure there's any guarantee that will remain true for future versions. The other upside of locking the Node.js version is (hopefully) fewer merge conflicts or unnecessary changes to the package-lock file as its format did change between the different versions of npm bundled with 7.5.0 and 10.15.3.

PR #189, opened 2019-05-03
Collect the references item that contains only href and text

cooltey:
This PR makes a V2 version of `collectNearbyReferences()`, and it will contain only the `href` value of the references and also the `text` instead of directly getting the `html` of the references from the web. 

If we get the `href` from the references group, we can easily map the data in the `references` endpoint.

https://phabricator.wikimedia.org/T221665


src/transform/ReferenceCollection.js
```diff
@@ -225,8 +266,23 @@ const collectNearbyReferences = (document, sourceNode) => {
   return new NearbyReferences(selectedIndex, referencesGroup)
 }
 
+/**
+ * Collect nearby references.
+ * @param {!Document} document
+ * @param {!Node} sourceNode
+ * @return {!NearbyReferencesV2}
+ */
+const collectNearbyReferencesV2 = (document, sourceNode) => {
```
montehurd: May be nice to have a more declarative name (`V2` isn't very informative 😂) maybe `collectNearbyReferencesAsText`?

cooltey: Sounds good! Done

montehurd: I'm still seeing `V2`... is there a way you could describe what's different about this new class and name it accordingly?

cooltey: Changed from `ReferenceItemV2` to `ReferenceLinkItem` since it contains the `href` and `text` inside a `<a>` tag. Not sure if it works for you as well?

PR #190, opened 2019-05-13
Fix collapsed infobox float

mateusbs17:
Collapsed infobox is floating to the right due to parent rule.

Bug: [T221695](https://phabricator.wikimedia.org/T221695)

### DIFF
--- a/src/transform/CollapseTable.css
+++ b/src/transform/CollapseTable.css
@@ -190,3 +190,9 @@ div.pagelib_collapse_table_container {
     float: initial;
   }
 }
+
+/* Fix Collapsed infobox floating to the right, see https://phabricator.wikimedia.org/T221695
+   .mw-stack.mobile-float-reset is a parent element to the infobox table*/
+.mw-stack.mobile-float-reset {
+    float: none !important;
+}


mateusbs17:
@montehurd and @berndsi not sure if that is the best approach to the problem, `.mw-stack.mobile-float-reset` seems to be a too general rule. 

It would be nice to have something like the `:has()` selector in JS, but I couldn't find anything similar to CSS.

What do you think?

berndsi:
Here are my thought (not really blockers):

Yeah, unfortunately CSS `:has()` is not supported by browsers yet. Not a big fan of using `!important` in CSS rules but it's tough to break inline style specificity otherwise.

The inline style for this comes from https://en.wikipedia.org/wiki/Module:Infobox_military_conflict. I'm wondering if there's a CSS class that this template should use instead of inline styles that would behave as expected on desktop, and ideally on mobile, too. Maybe we wouldn't even need to add additional CSS classes to pagelib in that case? 
Then I wonder how many other Lua modules/templates are affected by this and what their class names would be to select those.

joewalsh:
@mateusbs17 It looks like [mobile.css](https://en.wikipedia.org/wiki/MediaWiki:Mobile.css) abandoned overriding `mobile-float-reset` in favor of trying to get templates to adopt TemplateStyles.

Since we're altering the DOM already for collapsed tables, it seems we should remove any parent elements between the table and the `content_block` that might alter the style. I tried that approach here: https://github.com/wikimedia/wikimedia-page-library/pull/196

mateusbs17:
@joewalsh thanks for the heads up. I'll follow up in the PR you created.

PR #192, opened 2019-05-22
Unset any explicit height of collapsible tables

joewalsh:
When adding `display: block` any height mistakenly set in the table CSS is applied, causing the table to be truncated.

https://phabricator.wikimedia.org/T223477

### DIFF
--- a/src/transform/CollapseTable.css
+++ b/src/transform/CollapseTable.css
@@ -180,7 +180,7 @@ div.pagelib_collapse_table_container {
 .pagelib_collapse_table_container table {
     display: block;
     width: 100% !important;
-    height: unset !important;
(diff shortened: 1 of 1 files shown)


src/transform/CollapseTable.css
```diff
@@ -179,7 +179,8 @@ div.pagelib_collapse_table_container {
    but only for tables we collapse. */
 .pagelib_collapse_table_container table {
     display: block;
-    width: 100% !important
+    width: 100% !important;
+    height: unset !important;
```
montehurd: I think Android had a compatibility issue with `unset` and `initial` was preferred... https://github.com/wikimedia/wikimedia-page-library/pull/176

PR #194, opened 2019-06-03
Bump patch versions of dependencies to resolve security issues

mdholloway:
Bumped a couple of dependencies to the latest patch versions identified
by npm audit with `npm audit fix`.

A vulnerability remains in the axios package, which is an indirect
dependency via two direct dependencies. Neither direct dependency has
yet updated to the fixed release.


mdholloway:
@joewalsh noticed that this adds a bunch of `"optional": true` lines to package.lock.json.  I think this is because I'm on a Linux system, because our node/npm setups are otherwise identical.  I'll close this, but someone on a Mac should create a similar patch.

berndsi:
see also my comment in #200

PR #196, opened 2019-06-05
Ensure collapsed tables are added to content_block divs, eliminating any parentNodes in between

joewalsh:
Fix for https://phabricator.wikimedia.org/T221695. 

To standardize the appearance of collapsed tables, we should probably sure there are no additional elements between the `pagelib_collapse_table_container` div and the `content_block`. In this case, the [module for the military conflict infobox](https://en.wikipedia.org/wiki/Module:Infobox_military_conflict) adds a div with an inline style of `float:right` around the infobox.

**Before fix:**
Simulator Screen Shot - iPhone X - 2019-06-05 at 14 19 32
Simulator Screen Shot - iPhone X - 2019-06-05 at 14 19 36

**After fix:**
Simulator Screen Shot - iPhone X - 2019-06-05 at 14 18 46
Simulator Screen Shot - iPhone X - 2019-06-05 at 14 18 42


berndsi:
I think it would be nice to add a unit test for this.

src/transform/CollapseTable.js
```diff
@@ -309,6 +309,37 @@ const newCaptionFragment = (document, title, headerText) => {
...
+  if (!contentBlock) {
+    return
+  }
+  let foundContentBlock = false
+  while (contentBlock) {
+    if (contentBlock.classList && contentBlock.classList.contains('content_block')) {
+      foundContentBlock = true
+      break
+    }
+    lastAncestorBeforeContentBlock = contentBlock
```
berndsi: Since this is going upwards, isn't `lastAncestorBeforeContentBlock` really `childOfContentBlock`?

joewalsh: Updated

joewalsh:
Added a check for `data-mw-section-id` to fix compatibility with mobile-html. Verified it fixes the issue when running locally.

joewalsh:
@berndsi this should be good to go and is working with mobile-html

test/transform/CollapseTable.test.js
```diff
@@ -838,6 +838,44 @@ describe('CollapseTable', () => {
...
+      assert.ok(collapsibleContainer)
+      assert.ok(collapsibleContainer.classList.contains('pagelib_collapse_table_container'))
+      const table = collapsibleContainer.querySelector('.infobox')
+      assert.ok(table)
+      assert.deepEqual(table.parentNode, collapsibleContainer)
+    })
+
+    it('moves a table to the tag with the data-mw-section-id attribute when it is nested', () => {
+      const html = `
+        <div><section data-mw-section-id=0><div class=not_content_block><div>
```
berndsi: To make this test more realistic: Parsoid doesn't have an outer div around the `<section>` tag.

joewalsh: Fixed

PR #197, opened 2019-06-10
page library: allow clients to adjust text size

mateusbs17:
When native clients load articles via mobile-html, they need to be able
to adjust the font size.

PS: AllTransforms demo coming in a separate patch with some refactoring.
Bug: T222752

### DIFF
--- a/src/pcs/c1/PageMods.js
+++ b/src/pcs/c1/PageMods.js
@@ -1,3 +1,4 @@
+import AdjustTextSize from '../../transform/AdjustTextSize'
 import BodySpacingTransform from '../../transform/BodySpacingTransform'
 import CollapseTable from '../../transform/CollapseTable'
 import DimImagesTransform from '../../transform/DimImagesTransform'
@@ -55,6 +56,12 @@ const setMulti = (document, settings, onSuccess) => {
   }
   if (settings.scrollTop !== undefined) {
     Scroller.setScrollTop(settings.scrollTop)
+  }
+  if (settings.setTextSizeAdjustmentPercentage !== undefined) {
+    AdjustTextSize.setTextSizeAdjustmentPercentage(
+      document.body,
+      settings.setTextSizeAdjustmentPercentage
+    )
   }
 
   if (onSuccess instanceof Function) {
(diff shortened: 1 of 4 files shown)


mateusbs17:
@berndsi this is ready for review. 

Note that I would like your review on https://github.com/wikimedia/wikimedia-page-library/pull/198 before adding the new transformation to the all transforms demo. Let me know if you have any questions or concerns.

mateusbs17:
@berndsi it is ready for review but doesn't include an All Transformations option because I'm waiting for your opinion about https://github.com/wikimedia/wikimedia-page-library/pull/198.

Let me know if you have any questions or concerns.

src/pcs/c1/PageMods.js
```diff
@@ -56,6 +57,12 @@ const setMulti = (document, settings, onSuccess) => {
   if (settings.scrollTop !== undefined) {
     Scroller.setScrollTop(settings.scrollTop)
   }
+  if (settings.setTextSizeAdjustmentPercentage !== undefined) {
```
berndsi: For consistency sake let's drop the `set` prefix and call this just `textSizeAdjustmentPercentage`.
Also add this to the list of fields in the comment on line 37.

test/transform/AdjustTextSizeTransform.test.js
```diff
@@ -0,0 +1,20 @@
...
+      const document = domino.createDocument('<p></p>')
+      AdjustTextSize.setPercentage(document.body)
+      assert.strictEqual(document.body.outerHTML, '<body><p></p></body>')
+    })
+
+    it('set string value', () => {
+      const document = domino.createDocument('<p></p>')
+      AdjustTextSize.setPercentage(document.body, '80%')
+      assert.strictEqual(document.body.style['-webkit-text-size-adjust'],'80%')
+      assert.strictEqual(document.body.style['text-size-adjust'],'80%')
```
berndsi: missing spaces after comma in lines 16 and 17

src/pcs/c1/PageMods.js
mateusbs17: Done.

mateusbs17:
@berndsi now ready for review. I noticed that All transformations demo uses meta.wiki pagelib. Is that right? This means the demo I inserted can't be tested until some version is deployed?

berndsi:
I'm ok leaving the demo as part of a separate PR. I think we should consider using localhost for it if we can. Not sure how easy it is, though.

mateusbs17:
> I'm ok leaving the demo as part of a separate PR. I think we should consider using localhost for it if we can. Not sure how easy it is, though.

Done. I will wait until this is patch is OK to submit the demo changes.

berndsi:
Looks good to me. (Not sure where the approve button went.)

PR #200, opened 2019-06-11
update package lock file

montehurd:


berndsi:
Looks to be the new version of npm (6.6?) https://stackoverflow.com/questions/54411377/package-lock-json-file-package-with-optional-true

joewalsh:
@montehurd @berndsi this puts the `package-lock.json` file out of sync with the npm version (6.4.1) bundled with the node version (10.15.3) specified by the repo. In the future, please use the specified version or also make the corresponding update to the `.node-version` file.

berndsi:
> @montehurd @berndsi this puts the `package-lock.json` file out of sync with the npm version (6.4.1) bundled with the node version (10.15.3) specified by the repo. In the future, please use the specified version or also make the corresponding update to the `.node-version` file.

Sorry for not catching that. It's probably because we were not using node-env.

PR #201, opened 2019-06-11
TemplateStyles CSS appearing in quick facts header

mateusbs17:
Per https://phabricator.wikimedia.org/T215059#4933826, is not possible to use innerText, this patch is stripping style and script from fragmentHeader instead.

Bug: [T215059](https://phabricator.wikimedia.org/T215059)


berndsi:
Code change looks good. Would be good to add a unit test, though.

mateusbs17:
> Code change looks good. Would be good to add a unit test, though.

Done.

PR #215, opened 2019-07-11
Rename areTablesCollapsed to areTablesInitiallyExpanded

berndsi:
The old name was misleading and lead to T227783.

Under the covers we only toggle the collapsed/expanded state for all
tables, which is only useful right when the page gets loaded before the
user makes any changes, hence the addition of "Initially".
The default is all tables are collapsed, so if the user wants the tables
expanded areTablesInitiallyExpanded should be set to true.

Bug: https://phabricator.wikimedia.org/T227783


thesocialdev:
LGTM.

PR #217, opened 2019-07-12
Fix: update the article description text styles from px to em

cooltey:
Didn't see it happen on the iOS app but in the Android app, updating the font size will not change the article description text size.

ezgif com-resize


montehurd:
Just curious, on Android how are you adjusting the size css?

On iOS there's a single css selector we can use which proportionally scales everything (including spacing) regardless of units `em`/`px` etc.

https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html#//apple_ref/doc/uid/TP40006510-SW16

Wonder if the standard version https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust would work for Android too?

cooltey:
> 
> 
> Just curious, on Android how are you adjusting the size css?
> 
> On iOS there's a single css [selector](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html#//apple_ref/doc/uid/TP40006510-SW16) we can [use](https://github.com/wikimedia/wikipedia-ios/blob/develop/Wikipedia/Code/WKWebView%2BWMFWebViewControllerJavascript.m#L12) which proportionally scales everything (including spacing and images) regardless of units `em`/`px` etc.
> 
> Wonder if the standard version of the [selector](https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust) would work for Android's minimum OS target too?

In Android we use [setDefaultFontSize()](https://developer.android.com/reference/android/webkit/WebSettings#setDefaultFontSize(int)) to set the font size. 

Ideally, it should work on the `px` as well, but I am not sure why it does not work in this specific case. Maybe Google has changed the logic of adjusting text size in the WebView component (since it can be updated through Google Play).

I think the selector would work for the Android's minimum OS target.

montehurd:
Does `setTextZoom` work?

cooltey:
> 
> 
> Does `setTextZoom` work?

Thanks, @montehurd!
Not sure why we didn't use it before, maybe it is because it requires `minSdkVersion = 14`. Since we increase it to `minSdkVersion = 21`, I think we can have a try on this.

cooltey:
It can be fixed on Android client side.

cooltey:
Reopen this PR due to the feedback from: https://phabricator.wikimedia.org/T230635

PR #218, opened 2019-07-12
mobile-html: DOMRect object from InteractionHandler response cannot be decoded on iOS

nambatee:
https://phabricator.wikimedia.org/T227907

I'm assuming this will work on Android since plain objects are passed in other places

Let me know if there's a better way to do this 🤔

### DIFF
--- a/src/transform/ReferenceCollection.js
+++ b/src/transform/ReferenceCollection.js
@@ -122,6 +122,25 @@ class ReferenceLinkItem {
 }
 
 /**
+ * Get node's bounding rect as a plain object.
+ * @param {!Node} node
+ * @return {!Object<string, number>}
+ */
+const getBoundingClientRectAsPlainObject = node => {
+  const rect = node.getBoundingClientRect()
+  return {
+    top: rect.top,
+    right: rect.right,
+    bottom: rect.bottom,
+    left: rect.left,
+    width: rect.width,
+    height: rect.height,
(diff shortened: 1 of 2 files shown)


thesocialdev:
LGTM.

berndsi:
@nambatee We could also just revert this commit: https://github.com/wikimedia/wikimedia-page-library/commit/0c247ca19c7bbdb87a8a866e909aecbba02ade5e

joewalsh:
@berndsi I don't think 0c247ca should be reverted - the comment there is misleading. This isn't just for compatibility with `CGRect`, it's to allow it to be serialized by iOS at all by converting it to a plain JS object

berndsi:
Either way the comments should be updated to make this clearer. It's still unclear to me why omitting a few properties in JSON makes the iOS app crash.

#218 vs reverting 0c247ca  is mainly about where to put this functionality. If we merge #218 then the underlying `ReferenceCollection.js` will change. But ok, adding amore fields should not cause any issues for current pagelib clients.

joewalsh:
@berndsi it's not about omitting properties - it's about making it a plain JS object instead of a `DOMRect`. iOS is unable to serialize the `DOMRect` and pass it to our native code, but it can serialize a plain JS object and pass it along

berndsi:
Sorry it took a bit but I understand now. Thank you for your patience!

PR #221, opened 2019-07-23
Add href to ReferenceCollection to collectNearbyReferences output

berndsi:
It's useful to know what the target link to the reference is for the
reference_clicked event in InteractionHandling.


berndsi:
Started this to address an issue @Cooltey raised.

cooltey:
Thanks @berndsi for having this PR.

Same reason from #189, if we can get `href` from page-library, we can use the filtered `href` value to find the reference data from the `api/rest_v1/page/references/`.

PR #222, opened 2019-07-24
Pass href with imageClicked action, don't pass natural size. Pass title and text for linkClicked actions

joewalsh:
- Fix for https://phabricator.wikimedia.org/T228536 , pass `href` similar to other actions. In this case it points to the original `File:` title
- Remove naturalWidth, naturalHeight - not sure this will be reliable and I don't think the clients need it, only the original width and height that's already included CC @sharvaniharan @nambatee 
- Fix for https://phabricator.wikimedia.org/T228535, pass `title` and `text` with `linkClicked`

It looks like the placeholder action doesn't work as it was (it sends a linkClicked event) - feel free to fix that here, otherwise it can be a follow on

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "bugs": "https://github.com/wikimedia/wikimedia-page-library/issues",
   "main": "build/wikimedia-page-library-transform.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


sharvaniharan:
That is correct @joewalsh naturalWidth, naturalHeight are not needed on Android.

PR #223, opened 2019-07-24
Add getSelectionInfo

joewalsh:
Add `getSelectionInfo` to `pagelib.c1.InteractionHandler`
https://phabricator.wikimedia.org/T228686

### DIFF
--- a/docs/pcs/pcs.md
+++ b/docs/pcs/pcs.md
@@ -213,7 +213,11 @@ pagelib.c1.InteractionHandling.getSelectionInfo()
 
 Should return something along the lines of:
 ```
-{text: "selected text here", section: "1", isTitleDescription: false}
+{
+  text: "selected text here",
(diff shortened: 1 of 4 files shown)


berndsi:
@joewalsh Thank you! This works well. It would be good to add documentation for this to the pcs.md file. Is the action/event for the text selection going to be a different PR?

joewalsh:
@berndsi I think the action/event may have been a miscommunication in the PR - it looked like in the android code it only needed to be a read of selection information, not an an action when selection changes - @sharvaniharan is this correct?

joewalsh:
@berndsi I updated the doc

berndsi:
@joewalsh I think you're correct. My bad. Should we add in the docs that `section` may be null (in the case of the selected text being outside of a section, e.g. in the description)?

joewalsh:
@berndsi done

PR #225, opened 2019-07-25
WIP: Prevent placeholder to be wider than the viewport width on initial load

thesocialdev:
Bug: [T228287](https://phabricator.wikimedia.org/T228287)

### DIFF
--- a/src/transform/WidenImage.css
+++ b/src/transform/WidenImage.css
@@ -72,5 +72,5 @@ div.gallerytext > p {
 
 /* Prevent image placeholder to be wider than the viewport */
 .gallerybox .pagelib_lazy_load_placeholder {
-  width: 100%!important;
+  width: 100% !important;
 }
--- a/src/transform/WidenImage.css
+++ b/src/transform/WidenImage.css
@@ -69,3 +69,8 @@ li.gallerybox div.thumb > div {
 div.gallerytext > p {
   margin-top: 0px !important;
 }
+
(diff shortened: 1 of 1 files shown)


src/transform/WidenImage.css
```diff
@@ -68,4 +68,9 @@ li.gallerybox div.thumb > div {
 }
 div.gallerytext > p {
   margin-top: 0px !important;
+}
+
+/* Prevent image placeholder to be wider than the viewport */
+.gallerybox .pagelib_lazy_load_placeholder {
+  width: 100%!important;
```
berndsi: Would you add a space between the `%` and the `!`?
Have you considered using `max-width` instead of `width`?

thesocialdev: > Would you add a space between the `%` and the `!`?

Done.
> Have you considered using `max-width` instead of `width`?

It's already applied https://github.com/wikimedia/wikimedia-page-library/blob/master/src/transform/LazyLoadTransform.css#L38

berndsi: Good catch! Thanks!

thesocialdev:
@berndsi and @joewalsh this PR might be unnecessary in favor of fixing a adjust image bug at the server side.

So, we can close it in favor of https://gerrit.wikimedia.org/r/#/c/mediawiki/services/mobileapps/+/525583

joewalsh:
I agree it should be fixed server-side as that also fixes the issue that caused the images to overflow

PR #236, opened 2019-07-30
pagelib: add References link to footer

thesocialdev:
When clicked it should post a message to the native side to allow it to
display the native component showing the reference lists.

Also, updated AllTransforms demo

Bug: [T228457](https://phabricator.wikimedia.org/T228457)


src/transform/FooterMenu.css
```diff
@@ -146,3 +146,7 @@ body[dir='rtl'] .pagelib_footer_menu_item_subtitle {
...
 }
+
+.pagelib_theme_sepia .pagelib_footer_menu_icon_reference_list {
+  background-image: url('');
```
thesocialdev: @berndsi and @joewalsh this one I don't know how to do it, could you please help me?

berndsi: Good question. Maybe @montehurd knows? Also: where do we get the icon for references from?

joewalsh: @thesocialdev it can remain empty for now or you can just copy one of the existing ones as a placeholder. The icons will all be updated after https://phabricator.wikimedia.org/T226094 is completed.

PR #240, opened 2019-08-01
Performance: Prefer firstChild/nextSibling to childNodes

mdholloway:
Per Domino's readme file,[1] we should prefer traversing the DOM tree
using The `first/lastChild` and `next/previousSibling` properties rather than
`childNodes`. Doing the latter converts the linked list representation of
the DOM tree structure to an array-based representation, hurting
performance when modifying the tree.

Updated tests that were accessing `childNodes[0]` rather than `firstChild`
as well, to reinforce this.

[1] https://github.com/fgnass/domino#optimization

### DIFF
--- a/src/transform/CollapseTable.js
+++ b/src/transform/CollapseTable.js
@@ -26,7 +26,7 @@ const CLASS = {
  * @return {!boolean}
  */
 const isHeaderEligible =
-  header => header.childNodes && Polyfill.querySelectorAll(header, 'a').length < 3
+  header => Polyfill.querySelectorAll(header, 'a').length < 3
 
 /**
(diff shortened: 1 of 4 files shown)


joewalsh:
@montehurd the performance improvement isn't arguable and that's the underlying problem causing issues when running these transforms within mobileapps services.

mdholloway:
MCS is currently falling over because it contains large swaths of code written with no attention to performance.

https://phabricator.wikimedia.org/T229286

This is an intervention that produced noticeable improvements earlier in MCS.

montehurd:
@joewalsh 

> that's the underlying problem causing issues when running these transforms within mobileapps services.

^ good argument :)

PR #245, opened 2019-08-26
Update footer style

thesocialdev:
TODO:
- [x] Icons
- [x] Section segment placeholder
- [x] Section title font and style
- [x] Wikipedia Footer


thesocialdev:
This PR introduces: 
1) New styles for the available themes
2) Remove save buttons
3) Allow  clients to pass in page fragments

Default
image

Sepia
image

Black
image

Dark
image

Any feedback, question or concerns? cc// @berndsi, and @joewalsh

thesocialdev:
The drawback so far is CSS bundle size, which increased due to the new base64 images.

thesocialdev:
@montehurd, @berndsi, and @joewalsh this is ready for review.

joewalsh:
@thesocialdev it looks like the background color doesn't match the mockup - [asked for clarification here](https://phabricator.wikimedia.org/T226094#5480141) because it looks like it'd affect the implementation

thesocialdev:
> @thesocialdev it looks like the background color doesn't match the mockup - [asked for clarification here](https://phabricator.wikimedia.org/T226094#5480141) because the answer would affect the implementation

I actually didn't understand your question. Are you referring to the width of the borders which are not edge to edge in the Ipad resolution?

For the background-color I was following the zepellin mockups [1] [2] provide by Carolyn, and for the theme variants, I looked into old designs to use an acceptable design since the footer mockups don't have theme variants in the mockups.

[1] https://zpl.io/aMRj6dd
[2] https://zpl.io/bJWgMQK

joewalsh:
@thesocialdev I am referring to the background color behind the headers and the copyright footer that should be different than the background color behind the menu items and read more items:
Screen Shot 2019-09-16 at 2 01 21 PM

The [question on the ticket](https://phabricator.wikimedia.org/T226094#5480141) is on whether or not the menu items and read more items should extend the full width of the page or not.

joewalsh:
It looks like the top border is missing for the initial header - could the top and bottom border move to `pagelib_footer_container_menu_heading` instead of `pagelib_footer_container_menu_items` and  `pagelib_footer_container_readmore_pages`?

Screen Shot 2019-09-18 at 11 38 06 AM

thesocialdev:
> It looks like the top border is missing for the initial header - could the top and bottom border move to `pagelib_footer_container_menu_heading` instead of `pagelib_footer_container_menu_items` and `pagelib_footer_container_readmore_pages`?

Done.

PR #246, opened 2019-08-27
Use a shell page which loads the page content inline

berndsi:
Started also a load function but it's not used yet.


berndsi:
Abandoned in favor of #247.

PR #252, opened 2019-10-03
Fix: Get the section header element correctly

mdholloway:
Fixes an issue caused by a structural difference between raw Parsoid
HTML and PCS mobile-html. In Parsoid HTML, a section's heading element is
always the first child of a section element, but in mobile-html it's
wrapped in an enclosing div.


mdholloway:
Actually, please hold off merging, I'm going to update the tests too

PR #255, opened 2019-10-21
Add 'notheme'

joewalsh:
https://phabricator.wikimedia.org/T229286

`classifyElements` is by far the slowest part of preparing `mobile-html`. It's not sustainable to keep adding exclusions to the list of specific [filenames](https://github.com/wikimedia/wikimedia-page-library/blob/master/src/transform/ThemeTransform.js#L44-L45) and [query selectors](https://github.com/wikimedia/wikimedia-page-library/blob/master/src/transform/ThemeTransform.js#L83-L98). 

I'm proposing we add `notheme` and request editors add it to the appropriate templates ahead of the launch of `mobile-html`. This will allow the `mobile-html` endpoint to stop calling `classifyElements`. There is some precedent for this with `nomobile` (now obviated by TemplateStyles) but will need to confirm that this is a viable option.

### DIFF
--- a/src/pcs/c1/Page.js
+++ b/src/pcs/c1/Page.js
@@ -236,12 +236,17 @@ const getScroller = () => Scroller
  * @return {void}
  */
 const onBodyStart = () => {
+  if (!document) {
+    return
+  }
+
   // eslint-disable-next-line no-undef
   if (typeof pcsClient !== 'undefined' && pcsClient.getSetupSettings) {
     // eslint-disable-next-line no-undef
     const setupJSON = pcsClient.getSetupSettings()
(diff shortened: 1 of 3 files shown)


thesocialdev:
Does it still need to be confirmed as a viable solution? Waiting on that to merge this PR and https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/544854

joewalsh:
@thesocialdev It still needs to be confirmed that this is a viable solution and there's not a better option. Adding the `hold` tag for now. https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/544854 can be merged independently.

